PR #3, opened 2015-11-04
`this` expressions

### LINKED ISSUE #2: `this` parses as an identifier
I feel like it should probably be categorized in its own right. At the very least this would allow it to be syntax highlighted differently.

robrix:
Fixes #2.

I named the production `this_expression` to [match ESTree’s spec](https://github.com/estree/estree/blob/master/spec.md#thisexpression).

I haven’t added `arguments` in this PR since a) there’s no such production in ESTree & I’m unsure how faithful we want to be to that, and b) I wanted to be sure I’d got this right first :blush:

### DIFF
--- a/grammar.coffee
+++ b/grammar.coffee
@@ -322,6 +322,8 @@ module.exports = grammar
         /\d+/,
         optional(seq(".", /\d*/)))))
 
+    this_expression: -> "this"
+
     identifier: -> /[\a_$][\a\d_$]*/
 
     true: -> "true"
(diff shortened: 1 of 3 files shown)


robrix:
screen shot 2015-11-04 at 10 30 24

Wow! I guess that’s pretty characteristic of grammar changes since `parser.c` gets (essentially) rewritten.

maxbrunsfeld:
Yeah, sounds like there's no need to add `arguments` then. Thanks for the fix!

robrix:
:bow: thanks for the merge! :blush:

PR #8, opened 2016-05-25
Optional terminators

joshvera:
At the moment the javascript grammar fails to parse javascript with optional semicolons, and there doesn’t seem to be a way to [insert semicolons](http://www.ecma-international.org/ecma-262/6.0/#sec-rules-of-automatic-semicolon-insertion) when we run into offending tokens. @maxbrunsfeld any ideas?

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -60,21 +60,19 @@ module.exports = grammar({
       $.throw_statement
     ),
 
-    expression_statement: $ => seq(
-      err(choice($._expression, $.comma_op)), terminator()
-    ),
+    expression_statement: $ =>
+      err(choice($._expression, $.comma_op)),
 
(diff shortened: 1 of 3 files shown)


maxbrunsfeld:
Yeah, optional semicolons are currently supported by allowing a `\n` to function as a terminator. This doesn't apply to the last statement in a one-line block though, as shown by your unit test. I think it shouldn't be too tricky to make that case work.

It looks like you already took steps in exactly the right direction - moving the terminator token outside of the individual statement rules. I see there's some LR conflicts in the grammar now. Let me take a :eyes:...

maxbrunsfeld:
@joshvera You ok w/ closing this out?

joshvera:
👍

PR #9, opened 2016-05-26
Don't require a semicolon or line-break after the last statement in a block/program

maxbrunsfeld:
Refs #8 

This solves the optional semicolon problem in a different way than @robrix [used](https://github.com/tree-sitter/tree-sitter-ruby/blob/ce1cc836dfaf21d0e8d44ed58c49780f1d53b9ff/grammar.js#L49) in the ruby grammar. Rob treated the `;` as a separator *between* statements, rather than treating it as *part* of the statement. 

What I did instead was to make it explicit that the *last* statement in a block is not required to end with a `;` or a `\n`. So the expression statement with no `;` in @joshvera's unit test parses as a `trailing_expression_statement`.

Rob's solution is more elegant, and in the case of ruby, better in every way. But optional semicolons are a little hairier in javascript than in ruby, because javascript allows code like this:

```js
foo      // <--- not the end of a statement
  .bar
  .baz;
```

Basically, in order to tell whether or not the first line-break functions as a terminator, you have to look forward to the next token. Ruby and Go (and I think Rust) disallow this, but in JS, it's super common 😬.

Unfortunately, tree-sitter isn't yet smart enough to handle code like the example above if we use Rob's technique. The notion of having tokens like `\n` that can either be syntactically meaningful or function as whitespace is something that I kind of made up for tree-sitter, rather than basing it on some established method. I tried to make it Just Work™, but it's not fully baked 😢.

In the meantime, do you think this solution will suffice?

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -34,7 +34,18 @@ module.exports = grammar({
   ],
 
   rules: {
-    program: $ => repeat($._statement),
+    program: $ => optional($._statements),
+
+    _statements: $ => choice(
+      seq($._statement, optional($._statements)),
+      choice(
+        $.trailing_break_statement,
+        $.trailing_yield_statement,
+        $.trailing_throw_statement,
+        $.trailing_return_statement,
+        $.trailing_expression_statement
+      )
+    ),
 
     //
     // Statements
@@ -64,6 +75,10 @@ module.exports = grammar({
       err(choice($._expression, $.comma_op)), terminator()
     ),
 
+    trailing_expression_statement: $ => seq(
+      choice($._expression, $.comma_op)
+    ),
+
     var_declaration: $ => seq(
       variableType(),
       commaSep1(err(choice(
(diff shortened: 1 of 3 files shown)


maxbrunsfeld:
This might be another argument in favor of @robrix's suggestion (https://github.com/tree-sitter/tree-sitter-ruby/issues/6) of allowing nodes to be somehow 'renamed' via the grammar, for semantic reasons. In this case, we don't *really* want separate names for `trailing_return_statement` than the normal `return_statement`.

Maybe something like:

```js
grammar({
  name: 'javascript',

  // proposed
  variants: {
    trailing_return_statement: 'return_statement'
  },

  rules: {
    // ...
  }
});
```

robrix:
> In the meantime, do you think this solution will suffice?

Sounds reasonable to me :+1:

joshvera:
Thanks for this @maxbrunsfeld! I see you modified grammar.json directly. I was under the impression that grammar.js was compiled to grammar.json. What’s the relation between the two?

maxbrunsfeld:
Yeah, `grammar.json` is generated based on `grammar.js`. I just check it in for convenience.

maxbrunsfeld:
Actually, I've got a better API:

```js
psuedonyms: {
  trailing_expression_statement: 'expression_statement',
  trailing_return_statement: 'return_statement'
}
```

PR #14, opened 2016-07-29
Add support for getter, setter, and generator methods

joshvera:


maxbrunsfeld:
✨

src/grammar.json
```diff
@@ -2779,8 +2791,21 @@
...
+                },
+                {
+                  "type": "STRING",
+                  "value": "set"
+                },
+                {
+                  "type": "STRING",
+                  "value": "*"
+                }
+              ]
```
robrix: Are these changes going to be overwritten when we bump `grammar.js`?

robrix: Never mind, my mistake 😆

PR #21, opened 2016-08-26
Parse async/await

joshvera:
🐇

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -393,6 +393,7 @@ module.exports = grammar({
     ),
 
     arrow_function: $ => seq(
+      optional('async'),
       choice(
         $.identifier,
         seq('(', optional($.formal_parameters), ')')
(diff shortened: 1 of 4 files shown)


joshvera:
🎁

maxbrunsfeld:
Nice. What do you think about doing `async` method definitions and arrow functions in the same PR?

maxbrunsfeld:
Apparently `async` can occur in object literals, not just classes: https://babeljs.io/repl/#?evaluate=false&lineWrap=false&presets=stage-0&code=obj%20%3D%20%7B%0A%20%20a%3A%20'b'%2C%0A%20%20async%20c()%20%7B%20'd'%20%7D%0A%7D

joshvera:
@maxbrunsfeld I think we should add it to method definitions, and object literals, but not arrow functions based off of https://github.com/tc39/ecmascript-asyncawait/issues/7. What do you think?

joshvera:
📫

maxbrunsfeld:
🚀  the future is now

deltab:
Why not arrow functions? They can be async too. See https://github.com/tc39/ecmascript-asyncawait/issues/7#issuecomment-196427529 and https://babeljs.io/repl/#?code=async%20()%20%3D%3E%201%0A.

maxbrunsfeld:
@deltab This PR did add support for async arrow functions.

Issue #22, opened 2016-08-26
Add destructuring support


joshvera:
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment.

PR #23, opened 2016-08-26
Add destructuring in assignments, var declarations, function parameters

maxbrunsfeld:
Fixes https://github.com/tree-sitter/tree-sitter-javascript/issues/22

See http://www.ecma-international.org/ecma-262/6.0/#sec-destructuring-assignment

I don't fully understand the productions containing 'initializer' in the es6 spec, and I haven't seen examples of that, so I'm going to hold off on modeling that for now. I think this will prevent another big set of errors when parsing ES6 files.

/cc @joshvera

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -30,7 +30,16 @@ module.exports = grammar({
 
   conflicts: $ => [
     [$._expression, $.method_definition],
-    [$._expression, $.formal_parameters]
+    [$._expression, $.formal_parameters],
+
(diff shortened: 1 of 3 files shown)


joshvera:
👍

maxbrunsfeld:
Ah just realized what `initializer` means; it's like a default value:

```js
{x, y = 5} = obj
```

I'll add those in a follow-up PR.

PR #25, opened 2016-10-05
Track tree-sitter api cleanup

tclem:
Track the API clean up that happened in https://github.com/tree-sitter/tree-sitter/pull/37.

@maxbrunsfeld, I was expecting `parser.c` to get regenerated too. Should running `npm build` do that? It's possible I just don't know how to regenerate. 😁  It's also possible that the only change to the parser is `TSDebugger -> TSLogger` and that that doesn't impact what we generate.

### DIFF
--- a/src/tree_sitter/parser.h
+++ b/src/tree_sitter/parser.h
@@ -43,7 +43,7 @@ typedef struct TSLexer {
   TSSymbol result_symbol;
 
   TSInput input;
-  TSDebugger debugger;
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
> I was expecting parser.c to get regenerated too.

I think I've already updated `parser.c` since the last time the format of the generated parsers changed, and those API renames are the only things that have changed in tree-sitter since then. So this is perfect.

:+1:

PR #26, opened 2016-11-23
JavaScript return grammar update

rewinfrey:
While debugging error productions, @tclem and I discovered the grammar was not as flexible as it should be for return statements.

Opening this as a place to track other possible grammar adjustments while working through more errors.

/cc @maxbrunsfeld @tclem @joshvera 

🎩 @tclem for the 🍐

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -272,7 +272,7 @@ module.exports = grammar({
 
     return_statement: $ => seq(
       'return',
-      choice($._expression, $.comma_op),
+      optional(choice($._expression, $.comma_op)),
       terminator()
     ),
 
--- a/grammar.js
+++ b/grammar.js
@@ -272,7 +272,7 @@ module.exports = grammar({
 
     return_statement: $ => seq(
       'return',
-      commaSep($._expression),
+      choice($._expression, $.comma_op),
       terminator()
     ),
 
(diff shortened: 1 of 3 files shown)


maxbrunsfeld:
Ah, thanks for catching that @rewinfrey!

I believe that we should parse this return value as a comma expression, rather than allowing a return statement to have multiple child expressions. In other places, I represent comma expressions using the [`comma_op`](https://github.com/tree-sitter/tree-sitter-javascript/blob/2628dc1558ec29f9962986081e2a82c9f40f7c6b/grammar.js#L555) rule, (which I've realized should probably renamed to `sequence_expression` to match the name in the [ESTree spec](https://github.com/estree/estree/blob/master/es5.md#sequenceexpression).

rewinfrey:
> I believe that we should parse this return value as a comma expression, rather than allowing a return statement to have multiple child expressions. In other places, I represent comma expressions using the comma_op rule, (which I've realized should probably renamed to sequence_expression to match the name in the ESTree spec.

👍 thanks for the explanation @maxbrunsfeld. I've updated the grammar to use `choice` and `comma_op` based on how `comma_op` was used elsewhere. Does that look correct to you?

maxbrunsfeld:
Great. Looks like we just need to update the tests to include a `comma_op`.

maxbrunsfeld:
🚢

PR #27, opened 2016-11-29
Support multiple function declarations

rewinfrey:
Our ability to parse multiple functions defined inline was problematic.

e.g.:

```javascript
function test(){function one(){}function two(){}function three(){}}
```

This patch makes it possible to parse multiple inline functions, although it does impact the way `expression_statement` parses `function`. Previously, every function was parsed as an individual `expression_statement`, but now, multiple sequential functions are defined as a single `expression_statement`. This approach may not be tenable, but figure it is a good place to get the ball rolling towards a solution.

/cc @maxbrunsfeld @tclem @joshvera

### DIFF
--- a/grammar_test/semicolon_insertion.txt
+++ b/grammar_test/semicolon_insertion.txt
@@ -61,7 +61,8 @@ function c() {return d}
 (program
   (expression_statement
     (function (identifier) (formal_parameters) (statement_block
-      (trailing_expression_statement (identifier))))
+      (trailing_expression_statement (identifier)))))
+  (expression_statement
     (function (identifier) (formal_parameters) (statement_block
       (trailing_return_statement (identifier))))))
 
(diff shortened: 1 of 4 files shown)


grammar_test/semicolon_insertion.txt
```diff
@@ -61,8 +61,7 @@ function c() {return d}
 (program
   (expression_statement
     (function (identifier) (formal_parameters) (statement_block
-      (trailing_expression_statement (identifier)))))
-  (expression_statement
```
rewinfrey: Fixture:

```javascript
function a() {b}
function c() {return d}
```

This patch means these functions are treated in sequence, and are regarded as a single `expression_statement`. Curious what @maxbrunsfeld thinks about this, and if this is an acceptable tradeoff for the relative small impact this change represents to the parser?

PR #30, opened 2017-02-10
Restructure some rules to allow typescript to extend this grammar

maxbrunsfeld:
⚠️ WIP ⚠️ 

See https://github.com/tree-sitter/tree-sitter-typescript

I think that [the Shift](http://shift-ast.org/parser.html) and the [ESTree](https://github.com/estree/estree) AST formats are probably the best guides for how we should structure this grammar, so if we're restructuring things, we should generally consult those.

* Extract a common `_pattern` rule that can be overridden by typescript. Should this be called `_binding` 😕 ?
* Rename `var_declaration` -> `variable_declaration`.
* Get rid of `var_assignment`, make every component of a variable declaration a `var_declarator` (whether it has an initializer or not).

/cc @joshvera Feel free to take this over or replace it with something completely different ⛵️ !

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -45,7 +45,7 @@ module.exports = grammar({
     // ( foo
     // ( foo
     //    ^-- arrow function parameter or parenthesized expression?
-    [$._formal_parameter, $._expression],
+    [$._pattern, $._expression],
 
     // ( {foo} )
     // ( [foo] )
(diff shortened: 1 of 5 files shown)


joshvera:
👍 this looks like a good start. I’ll take a look at those AST formats.

PR #32, opened 2017-02-24
Add class property fields

joshvera:


grammar.js
```diff
@@ -802,12 +802,18 @@ module.exports = grammar({
...
+        choice(
+          $.method_definition,
+          $._public_field_definition
+        ),
         optional(';')
       )),
       '}'
     ),
 
+    // TODO this should be a property_name
```
maxbrunsfeld: With this TODO, do you mean that we should rename this thing? Or are you talking about a bigger restructuring?

joshvera: Ah I mean this should be a larger restructuring. I believe the ES6 spec calls this a `PropertyName`.

maxbrunsfeld:
:+1:

PR #36, opened 2017-03-03
Add rest arguments to function calls

joshvera:


maxbrunsfeld:
I think we should give a name to the `seq('...', $.expression)`, like `rest_argument`, just that rest arguments are easy to identify from the tree.

joshvera:
👍

joshvera:
🏹

PR #38, opened 2017-03-09
Bump tree-sitter-cli for (foo - bar) / baz fix

joshvera:


grammar_test/expressions.txt
```diff
@@ -866,3 +866,15 @@ foo(...rest)
 (program
   (expression_statement
     (function_call (identifier) (arguments (rest_argument (identifier))))))
+
+==============================================
+Math expressions
+==============================================
+
+(foo - bar) / baz
```
maxbrunsfeld: Did this fail w/ the old one? I thought that for various reasons, the whole thing actually had to be in parentheses for it to fail.

joshvera: Yeah this failed with `0.4.0`.

PR #39, opened 2017-03-10
Allow nbsp as whitespace

joshvera:


grammar.js
```diff
@@ -26,7 +26,7 @@ module.exports = grammar({
   extras: $ => [
     $.comment,
     $._line_break,
-    /[ \t\r]/
+    /[ \t\r\uFEFF]/
```
tclem: What's this? unicode byte order mark?

joshvera: http://www.fileformat.info/info/unicode/char/feff/index.htm

maxbrunsfeld: This will probably come up in other languages. We should fix tree-sitter's implementation of the `\s` character class to include non-breaking spaces, since JavaScript's regexps seem to do this.

maxbrunsfeld: https://github.com/tree-sitter/tree-sitter/issues/67

PR #44, opened 2017-03-17
Add template string interpolation

joshvera:


grammar.js
```diff
@@ -656,9 +656,23 @@ module.exports = grammar({
...
-    )),
+    template_string: $ => seq(
+      '`',
+      repeat(choice(
+        $.template_chars,
+        $.template_substitution
+      )),
+      '`'),
+
+      template_chars: $ => prec.right(repeat1(
```
maxbrunsfeld: Can you try putting a `token` around this? I think that currently it will create a node for each character.

PR #45, opened 2017-03-18
Model automatic semicolon insertion using an external scanner

maxbrunsfeld:
This avoids the parse-stack branching that was previously happening when semicolons were omitted. It also remove `trailing_return_statement` and all of the other `trailing` rules. The parse state count has decreased from 3014 to 2441 and the parser generate time is down to about 12 seconds on my machine.

/cc @joshvera We'll have to figure out the best way to share the external scanner code between JavaScript and Typescript.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -23,10 +23,13 @@ const PREC = {
 module.exports = grammar({
   name: 'javascript',
 
+  externals: $ => [
+    $._automatic_semicolon
+  ],
+
   extras: $ => [
     $.comment,
-    $._line_break,
-    /[ \t\r\uFEFF\u2060\u200B]/
+    /[ \t\n\r\uFEFF\u2060\u200B]/
   ],
 
   conflicts: $ => [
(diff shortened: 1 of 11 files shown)


grammar.js
```diff
@@ -92,9 +91,14 @@ module.exports = grammar({
 
     // A function, generator, class, or variable declaration
     _declaration: $ => prec(PREC.DECLARATION, choice(
-      $.function,
-      $.generator_function,
-      $.class,
+      seq(
```
joshvera: @maxbrunsfeld Do you remember what this was for? Since we override `_declaration` in the typescript grammar, I'm not sure if I should bring this in or not.

maxbrunsfeld: Yeah, I added this was in order to make functions and classes parse as declarations rather than expression statements. When parsing this code:

```js
1| function foo() {}
2| var x
```

the lexer produces an `_automatic_semicolon` token at the end of line 1. Without this change, that token would cause *only* the `expression_statement` interpretation to be valid, not the `_declaration`.

PR #47, opened 2017-06-23
Grammar improvements

rewinfrey:
This patch adds improvements to the JavaScript grammar while working on https://github.com/tree-sitter/tree-sitter-typescript/pull/28. 

/cc @github/semantic-code @maxbrunsfeld

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -806,7 +806,7 @@ module.exports = grammar({
 
     _property_name: $ => choice($.identifier, $.reserved_identifier, $.string, $.number),
 
-    reserved_identifier: $ => choice('get', 'set', 'async', 'abstract', 'module', 'interface'),
+    reserved_identifier: $ => choice('get', 'set', 'async', 'abstract', 'interface'),
   }
 });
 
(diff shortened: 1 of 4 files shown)


grammar.js
```diff
@@ -775,7 +806,7 @@ module.exports = grammar({
 
     _property_name: $ => choice($.identifier, $.reserved_identifier, $.string, $.number),
 
-    reserved_identifier: $ => choice('get', 'set', 'async')
+    reserved_identifier: $ => choice('get', 'set', 'async', 'abstract', 'module', 'interface'),
```
maxbrunsfeld: Should these just be added in the typescript grammar? I'd think it'd be ok to just let these words be parsed as regular `identifiers` in regular JavaScript.

rewinfrey: @maxbrunsfeld I was basing this off this list: https://www.w3schools.com/js/js_reserved.asp. I just realized though that `module` is not a JavaScript keyword. I'll move `module` over to the TypeScript grammar, but I think `abstract` and `interface` are good here. What do you think?

rewinfrey:
@maxbrunsfeld thanks for the comment. I updated the grammar to remove `module` from the `reserved_identifier` list, and added `module` to the TypeScript grammar's `reserved_identifier` list. 🙇

PR #48, opened 2017-06-27
External comma scanner

rewinfrey:
🎩  @maxbrunsfeld for identifying the fix!

This allows us to parse comma delimited expressions like:

```javascript
var chai = require('chai')
  , expect = chai.expect
  , Support   = require(__dirname + '/../support')
  , current   = Support.sequelize
  , Sequelize = Support.Sequelize
  , sinon     = require('sinon');
```

/cc @maxbrunsfeld @joshvera

### DIFF
--- a/corpus/expressions.txt
+++ b/corpus/expressions.txt
@@ -95,6 +95,22 @@ $_;
   (expression_statement (identifier)))
 
 ============================================
+Multi-line variable declarations
+============================================
+
+var a = b
+  , c = d
+  , e = f;
+
+---
+
+(program
+  (variable_declaration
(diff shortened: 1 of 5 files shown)


maxbrunsfeld:
✨  Will you add a test for this, similar to what you posted in the PR description?

rewinfrey:
@maxbrunsfeld oh yeah! Thanks for catching that 🙇  Added in https://github.com/tree-sitter/tree-sitter-javascript/pull/48/commits/a5b81926fae8cc0be9fc296c49ab711ed763a8b4

PR #49, opened 2017-07-19
Use new grammar apis

maxbrunsfeld:
This PR updates the grammar to take advantage of tree-sitter's new [contextual renames feature](https://github.com/tree-sitter/tree-sitter/pull/90).

Some nodes that used to be simply called `identifier` will now be called `property_identifier`, `shorthand_property_identifier`, or `label_identifier`. Also, that pesky `reserved_identifier` rule is gone; when a keyword like `get` or `async` needs to be interpreted as a `property_identifier` or an `identifier`, it will appear exactly like any other node of that type.

Another improvement is that array and object destructuring patterns are now called `array_pattern` and `object_pattern` respectively, instead of appearing as an `array` or `object` wrapped in a `destructuring_pattern` node.

These changes will make the AST much easier to analyze in order to e.g. highlight all usages of a local variable in Atom.

### DIFF
--- a/corpus/destructuring.txt
+++ b/corpus/destructuring.txt
@@ -11,24 +11,24 @@ const {a, b: {c, d}} = object
 (program
   (expression_statement (assignment
     (object_pattern
-      (shorthand_property_name)
-      (shorthand_property_name))
-    (variable_name)))
+      (shorthand_property_identifier)
+      (shorthand_property_identifier))
+    (identifier)))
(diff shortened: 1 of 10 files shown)


rewinfrey:
👍  b/c this represents a potentially breaking change for the named nodes of the parse tree output, what do you think about a minor version bump for `tree-sitter-javascript` in `package.json`?

PR #54, opened 2017-09-19
Allow jsx expressions in jsx elements

joshvera:
It seems like `var b = <Text {...j}></Text>` is valid.


maxbrunsfeld:
:+1: Thanks!

PR #55, opened 2017-09-20
Add support for computed property names

joshvera:
To parse things like:

```javascript
{
  [methodNameStats]() {
  }
}
```


maxbrunsfeld:
Oh wow thanks, I can't believe this wasn't supported. I think that `computed_property_name` should probably be made visible, since there's such a big semantic difference between `{foo: bar}` and `{[foo]: bar}`.

joshvera:
:+1:

PR #60, opened 2017-10-06
Add rest parameters

joshvera:
- [x] Do statements can have single statements
- [x] Allow commas at the end of export clauses
- [x] Allow rest parameters in formal parameters


maxbrunsfeld:
Thanks!

PR #61, opened 2017-10-06
Disambiguate multiline comment and regex

joshvera:
The following code’s comment was parsing as a regex, so this PR tries to disambiguate that by saying regex opening characters can’t be followed by a `*`. /cc @maxbrunsfeld 

```javascript
firstNonWhitespaceIndex === -1
  ? /* entire string is whitespace */lineContent.length + 1
  : firstNonWhitespaceIndex + 1

```


maxbrunsfeld:
Ah that makes sense; in this case a `regex` and a `comment` would both be valid and both match the same length, so it's probably picking `regex` because it occurs earlier in the grammar. This seems like the right solution.

maxbrunsfeld:
Looks like tests are failing for a regex that starts with a character class, because we consume the opening `[` of the character class with the `[^*]` that you added.

Another solution you could try is putting a precedence on `comment`, inside of the `token` call. This will make the lexer prefer a comment over other tokens that match the same content.

```js
comment: $ => token(prec(1, seq(
  // ...
)))
```

maxbrunsfeld:
✨

PR #65, opened 2017-11-02
Move bitshift assignment operators to augmented assignment

joshvera:


maxbrunsfeld:
TIL! Thanks @joshvera!

PR #68, opened 2018-02-14
Improve JSX parsing

joshvera:
Follows https://facebook.github.io/jsx/ a little bit better.

- [x] Added JSXIdentifier
- [x] Added JSXNamespacedName
- [x] Added JSXFragment

### DIFF
--- a/corpus/expressions.txt
+++ b/corpus/expressions.txt
@@ -960,7 +960,7 @@ typeof a == b && c instanceof d
 Simple JSX elements
 ==============================================
 
-a = <div className='b' tabIndex=1 />;
+a = <div className='b' tabIndex={1} />;
 b = <Foo.Bar>a <span>b</span> c</Foo.Bar>;
 
 ---
@@ -970,19 +970,19 @@ b = <Foo.Bar>a <span>b</span> c</Foo.Bar>;
     (identifier)
     (jsx_self_closing_element
       (identifier)
-      (jsx_attribute (property_identifier) (string))
-      (jsx_attribute (property_identifier) (number)))))
+      (jsx_attribute (identifier) (string))
+      (jsx_attribute (identifier) (jsx_expression (number))))))
   (expression_statement (assignment_expression
     (identifier)
     (jsx_element
-      (jsx_opening_element (nested_identifier (identifier) (identifier)))
+      (jsx_opening_element (jsx_member_expression (identifier) (identifier)))
       (jsx_text)
       (jsx_element
         (jsx_opening_element (identifier))
         (jsx_text)
         (jsx_closing_element (identifier)))
       (jsx_text)
-      (jsx_closing_element (nested_identifier (identifier) (identifier)))))))
+      (jsx_closing_element (jsx_member_expression (identifier) (identifier)))))))
 
 ==============================================
 Expressions within JSX elements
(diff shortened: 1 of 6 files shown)


grammar.js
```diff
@@ -48,7 +55,10 @@ module.exports = grammar({
     [$._expression, $.formal_parameters],
     [$.labeled_statement, $._property_name],
     [$.assignment_pattern, $.assignment_expression],
-    [$.computed_property_name, $.array]
+    [$.computed_property_name, $.array],
+    [$.jsx_identifier, $.jsx_attribute],
+    [$.jsx_identifier, $.jsx_namespace_name],
+    [$.jsx_identifier, $.jsx_member_expression]
```
maxbrunsfeld: Is there a way to make these fixes without introducing these conflicts? There might not be, but since each conflict carries some runtime cost, it'd be good to double check that we're modeling things in the most efficient way. Besides the naming changes, what new constructs are we parsing that we couldn't before?

corpus/expressions.txt
```diff
@@ -1086,3 +1086,21 @@ yield db.users.where('[endpoint+email]')
...
...dentifier) (jsx_self_closing_element (identifier) (jsx_attribute (identifier) (string)) (jsx_attribute (identifier) (jsx_expression (number))) (jsx_attribute (jsx_identifier (identifier) (identifier)) (string)) (jsx_attribute (identifier))))) (expression_statement (assignment_expression (identifier) (jsx_self_closing_element (identifier) (jsx_attribute (identifier) (jsx_expression (member_expression (identifier) (property_identifier))))))) (expression_statement (assignment_expression (identifier) (jsx_element (jsx_opening_element (identifier) (jsx_attribute (identifier) (jsx_expression (binary_expression (binary_expression (member_expression (member_expression (this) (property_identifier)) (property_identifier)) (string)) (unary_expression (member_expression (member_expression (this) (property_identifier)) (property_identifier))))))) (jsx_text) (jsx_closing_element (identifier))))) (expression_statement (assignment_expression (identifier) (jsx_element (jsx_opening_element (jsx_namespace_name (identifier) (identifier)) (jsx_attribute (identifier) (jsx_expression))) (jsx_expression (spread_element (identifier))) (jsx_closing_element (jsx_namespace_name (identifier) (identifier)))))))
```
maxbrunsfeld: This test is kinda hard to read w/ the single line. Can you format it a bit? We should really add some automated way to do it, I know it can be a bit tedious.

corpus/expressions.txt
```diff
@@ -1001,8 +1001,8 @@ b = <Text {...j}></Text>
     (identifier)
     (jsx_element
       (jsx_opening_element (identifier)
-        (jsx_attribute (property_identifier))
-        (jsx_attribute (property_identifier) (jsx_expression (identifier))))
+        (jsx_attribute (identifier))
+        (jsx_attribute (identifier) (jsx_expression (identifier))))
```
maxbrunsfeld: Ideally, I'd like for these property names to still be called `property_identifier`. It's just a naming thing, but currently, the grammar has a nice property that the identifiers are neatly categorized with regard to which ones are variable reference (`identifier`), which are names of object properties (`property_identifier`), which are type names `type_identifier`.

grammar.js
maxbrunsfeld: Ah, reading this a bit more, I can see that we're parsing JSX fragment syntax now. It'd be nice to have a test case that's specifically about that.

PR #73, opened 2018-04-16
Make example parsing not use submodules

tclem:
Running into some `Path too long` issues over the Windows build of https://github.com/tree-sitter/tree-sitter-typescript/pull/57. Try to parse examples without including example repos as submodules.

### DIFF
--- a/script/parse-examples
+++ b/script/parse-examples
@@ -4,14 +4,17 @@ set -e
 
 cd "$(dirname "$0")/.."
 
+
+# Clone npm/npm and check out a known sha
 repo=examples/npm
 
-if [ -d "$repo" ]; then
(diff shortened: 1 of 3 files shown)


script/parse-examples
```diff
@@ -2,12 +2,23 @@
 
 set -e
 
+cd "$(dirname "$0")/.."
+
+repo=examples/npm
+
+if [ -d "$repo" ]; then
+  pushd "$repo" && git pull
```
maxbrunsfeld: I was thinking maybe we should do `git reset --hard <SOME SPECIFIC SHA>` so that the build won't ever start breaking spontaneously because of changes they make on the master branch of `npm`. Thoughts?

PR #74, opened 2018-05-02
Print out some basic parse stats

tclem:
Gives us some basic stats on percentage of code that parses in particular example repositories.


tclem:
> Successfully parsed 99.58% of examples/npm (2603 of 2614 files)

🤘

PR #76, opened 2018-05-17
External template char

### LINKED ISSUE #67: Error when parsing a template string with a run of characters starting with `//`
For example, when parsing:

```js
const string = `${foo}//${bar}`
```

the scanner is producing a `comment` token when it sees the `//`. Once https://github.com/tree-sitter/tree-sitter/issues/111 is fixed, we'll be able to address this problem by putting a high precedence in the `template_chars` token.

314eter:
This fixes #67 because the external `_template_char` token has higher precedence than the internal `comment` token.


maxbrunsfeld:
👏  Great solution! Thanks for updating the known failures list too!

maxbrunsfeld:
Because [tree-sitter-typescript](https://github.com/tree-sitter/tree-sitter-typescript) extends this grammar, we'll need to duplicate the new external scanner logic in [the typescript external scanner](https://github.com/tree-sitter/tree-sitter-typescript/blob/master/src/scanner.c).

@314eter are you interested in doing that as well? I can take care of it otherwise.

maxbrunsfeld:
Oh, I just realized that this creates a token for every character in the string. I think this will have bad performance for long template strings. Could you adjust this so that there's a `template_chars` token that matches a continuous *series* of characters that are neither '$' nor '`'?

314eter:
I wanted to keep the scanner logic as simple as possible, but I didn't think about the performance.

PR #77, opened 2018-05-17
Scan series of template chars

314eter:
As suggested in https://github.com/tree-sitter/tree-sitter-javascript/pull/76#issuecomment-389997143.


maxbrunsfeld:
Looks great.

PR #89, opened 2018-12-07
Remove calls from new expressions

joshvera:
Constrain new expressions to `constructable_expression` and increase the precedence over call expressions.


maxbrunsfeld:
🤘 great improvement.

PR #90, opened 2018-12-07
Expose node for regex flags

pranaygp:
Expose the flags on a regex as its own node


maxbrunsfeld:
This is a great change. @pranaygp and I talked in slack about one small adjustment we should make - having the `regex_flags` node only appear if the regex has one or more flags.

pranaygp:
Updated 😄

maxbrunsfeld:
Excellent, thanks!

PR #95, opened 2019-01-08
Allow trailing commas in formal parameters

kevinsawicki:
This pull request allows trailing commas in formal parameters.

It also exposes a `_formal_paramter` rule that will be overridden in typescript since that grammar currently does not generate against the latest version of the JS grammar.

This was paired on with @tclem and @maxbrunsfeld 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas#Trailing_commas_in_functions


corpus/expressions.txt
```diff
@@ -478,7 +478,8 @@ Functions
   function(arg1, ...arg2) {
     arg2;
   },
-  function stuff() {}
+  function stuff() {},
+  function trailing(a,) {}
```
tclem: Is it worth testing more than 1 parameter here too?

kevinsawicki: Yeah, sounds good :+1:

PR #99, opened 2019-04-23
Fix relative precedence of plus and rel

marijnh:
It seems like `PREC.REL` should be lower than `PREC.PLUS` (since `1 + 1 < 10` parses like it does in C, `(1 + 1) < 10`).


maxbrunsfeld:
:+1: Thanks!

PR #100, opened 2019-07-11
Make grammar implementation repositories more friendly with tree-sitter Rust bindings

racer161:
Adding a cargo.toml, build.rs, and a lib.rs allows grammar definitions to be added to Rust projects using the tree-sitter Rust bindings easily by adding one dependency to cargo:

`[dependencies]`
`tree-sitter-javascript = { git = "https://github.com/racer161/tree-sitter-javascript" }` 

With this edition, the readme.md in tree-sitter/lib/binding_rust/ : [https://github.com/tree-sitter/tree-sitter/tree/master/lib/binding_rust](url) will compile without the need for a build.rs or any knowledge of the cc crate. There's also no need to upload these grammar repos to crates.io since they only expose C libraries.

Obviously, this would need to implemented for each grammar repository. I think this makes the most sense to foster support for the rust bindings since its a one time fix (most likely) without need for any further modification. I'd be willing to fork the Javascript, C and Rust repos as a trial so at least the readme in the rust bindings repo would work without knowledge of cc or building.


maxbrunsfeld:
@racer161 Sorry for ignoring this PR. It wasn't because of anything wrong with the PR. At the time I was just not prepared to make a decision about whether our team could maintain this kind of thing, and then lost track of this out of indecision.

We might move forward with something like this soon.

racer161:
> 
> 
> @racer161 Sorry for ignoring this PR. It wasn't because of anything wrong with the PR. At the time I was just not prepared to make a decision about whether our team could maintain this kind of thing, and then lost track of this out of indecision.
> 
> We might move forward with something like this soon.

No prob I totally understand its a pretty sweeping change. I've just been using my own fork for my personal projects and it works well enough. Thanks for the update.

maxbrunsfeld:
Ok, #141 implemented this.

PR #103, opened 2019-07-18
Fix for(x in a,b,c)

calixteman:
The right side of "in" operator in for statement can be a sequence.


maxbrunsfeld:
Nice! Thank you.

Issue #86, opened 2018-11-08
statement_block should have higher precedence than object


marijnh:
There's a test `{a, b} = object`, which I guess motivates the current precedence, but that's actually a syntax error—you (awkwardly enough) have to wrap statement-level destructuring assignments in parentheses in JavaScript, because an opening curly at statement level always starts a block.

maxbrunsfeld:
What construct do we parse incorrectly as a result? Stuff like `() => {}`, or `() => {aLabel: aCall()}`?

marijnh:
A block like `{}` or  `{foo: 2}` is treated as an object literal, for example.

maxbrunsfeld:
Fixed via #104.

PR #104, opened 2019-07-18
Empty bodies mustn't be an empty object

calixteman:
In expression for (...) { } the empty body is treated like an object because followed by an automatic_semicolon. So this expression is equivalent to to for (...) { }; and the empty object is finally an expression statement... which is of course wrong.
Anyway an object cannot be an expression_statement when it's the body, ideally we should forbid that with the grammar but it seems to be quite a pain.... so this workaround.


maxbrunsfeld:
Can you fix this just by changing precedence? See also https://github.com/tree-sitter/tree-sitter-javascript/issues/86. I don't think we need a separate token just for empty blocks.

calixteman:
I tried a lot of things and each one failed:
 - static pred,
 - dynamic pred with a conflict for statement_block and expression_statement
 - match the opening "{" in the scanner (but here I failed in consuming my 64 Gb of memory)
I've the impress that once the automatic_semicolon is guessed there is no way to have an other possibility...
But I'm not an expert, feel free to try and let me know what's the result.
As said, for (...) {"a":1} is valid in tree-sitter but in nodejs for example it's a failure... I don't care so much about that because I parse only valid code (I hope so :))

calixteman:
ah nice... I didn't think about prec.right.
TIL
thanks

maxbrunsfeld:
> each one failed

I think the problem you were seeing when using precedence is that after `{}`, an `_automatic_semicolon` token was valid because of the `(expression_statement (object))` interpretation. So the lexer returned an `_automatic_semicolon`. Then, upon seeing that token, the parser decided to reduce by `object -> '{'  '}'` because an automatic semicolon was not valid for a statement block.

PR #107, opened 2019-07-23
A label can have a cr before the colon

calixteman:
Because of automatic_semicolon, a label can't have a cr after it:
https://searchfox.org/mozilla-central/source/js/src/jsapi-tests/binast/parser/multipart/spidermonkey/ecma_2/Statements/label-003.js


maxbrunsfeld:
Good catch, Thanks!

PR #108, opened 2019-07-23
For in/of headers must have a lhs assignment expression as first expression

calixteman:
For reference:
https://www.ecma-international.org/ecma-262/8.0/#sec-for-in-and-for-of-statements


grammar.js
```diff
@@ -666,14 +667,16 @@ module.exports = grammar({
       '[', field('index', $._expressions), ']'
     )),
 
+    lhs_expression: $ => choice(
```
joshvera: Could we change this to `_lhs_expression` to hide this node in the output?

calixteman: Sure

PR #109, opened 2019-07-23
Catch can destructure its argument

calixteman:
For reference:
https://www.ecma-international.org/ecma-262/8.0/#prod-Catch


joshvera:
May need to regenerate `src/parser.c` here.

maxbrunsfeld:
Thanks!

PR #110, opened 2019-07-23
Import can be used as a function

calixteman:
For reference:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#Dynamic_Imports


maxbrunsfeld:
I think we should probably use `argument_list` in `import_expression`. This way, we will still handle any old JavaScript code in  which `import` was used as a regular function. Also, it should require fewer parse states.

calixteman:
I was thinking exactly the same when doing my supermarket shopping. I'll fix it.

PR #111, opened 2019-07-23
Reserved keyword can be a function parameter with a default value

calixteman:


grammar.js
```diff
@@ -819,8 +819,8 @@ module.exports = grammar({
       seq(
         '/*',
         /[^*]*\*+([^/*][^*]*\*+)*/,
-        '/'
-      )
+      '/'
+    )
```
maxbrunsfeld: I think the previous indentation was correct here. Was this intentional?

maxbrunsfeld:
Thanks so much for finding all these issues!

PR #112, opened 2019-07-23
Arg in arrow function can be a reserved keyword

calixteman:


corpus/expressions.txt
```diff
@@ -551,7 +552,9 @@ a => 1;
     (formal_parameters (identifier) (identifier))
     (number)))
   (expression_statement (arrow_function
-    (formal_parameters (identifier) (identifier)) (number))))
+    (formal_parameters (identifier) (identifier)) (number)))
+  (expression_statement (arrow_function
+    (binary_expression (identifier) (number)))))
```
maxbrunsfeld: There should be an additional `identifier` in here, for the `async` formal parameter. It looks like we just need an `alias` on the `_reserved_identifier` to make it appear as an identifier.

PR #114, opened 2019-07-24
LHS assignment expression can be parenthesized

calixteman:


maxbrunsfeld:
If you just use `parenthesized_expression` here, instead of having a special `_parenthesized_lhs_expression` rule, does it make the parser's state count larger or smaller?

calixteman:
@maxbrunsfeld, once you merge this PR, could you regenerate the parser for typescript please ?

maxbrunsfeld:
Done.

PR #115, opened 2019-07-24
An identifier can contain unicode escape sequence

calixteman:
For reference:
https://www.ecma-international.org/ecma-262/8.0/#prod-IdentifierName


maxbrunsfeld:
@calixteman just a heads up - I recently bumped the `tree-sitter-cli` dependency. Can you make sure you've got the latest `tree-sitter-cli` on these PRs?

calixteman:
No I used 0.15.5 and so I just updated it.
Since we need to regen parser between each PR merge, there is no need to regen everything, isn't it ?
Anyway, you could setup pre-commit: https://pre-commit.com/
It will allow to run generate/test in a commit hook and then we could check that ts version at least to say that a particular version must be used before generate.
Wdyt ?

maxbrunsfeld:
Wow, I was surprised to learn of that syntax. Thanks!

PR #117, opened 2019-07-24
Add import.meta as an expression

calixteman:
For reference:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import.meta


maxbrunsfeld:
After seeing this, I think that we should handle `import` more simply. Maybe we should remove the `import_expression` rule introduced in #110, and instead just add the `import` keyword as its own type of expression:

```js
_constructable_expression: $ => choice(
  $.identifier,
  $.this,
  $.import, // <-- new
  // ...
),

import: $ => 'import',
```

This way, `import("foo")` and `import.meta` would both parse using the normal rules `call_expression` and `member_expression` respectively, but you could still easily find all of the usages of `import` in a syntax tree by searching for `import` nodes. What do you think?

calixteman:
Yeah I think it's a way better like this.

maxbrunsfeld:
Thanks @calixteman!

PR #118, opened 2019-07-24
Underscore can be a numerical separator

calixteman:
For reference:
https://github.com/tc39/proposal-numeric-separator

It's implemented in firefox js engine.


maxbrunsfeld:
✨

PR #120, opened 2019-08-03
Update to 0.14.0-beta2 template

Aerijo:
Updates the `binding.cc` file to match the template on the Tree-sitter repo.

Fixes https://github.com/tree-sitter/tree-sitter/issues/409


maxbrunsfeld:
Thank you!

PR #121, opened 2019-08-09
A variable can be named set

calixteman:


calixteman:
Just made crappy things with git...

PR #122, opened 2019-08-09
A variable can be named set

calixteman:


calixteman:
@maxbrunsfeld ping ?

corpus/statements.txt
```diff
@@ -460,6 +460,7 @@ Variable declarations
 
 var x = 1;
 var x, y = {}, z;
+var set = 1;
```
maxbrunsfeld: I think this test would pass on master, because the `set` token is not valid in this context. Is there some case that you're seeing in the wild that does not parse successfully on master? It would be good to tweak this test to cover that case.

calixteman: It's ok with master so I just keep the test.

mjambon:
I'm taking the liberty of closing this two-year old pull request.

PR #136, opened 2020-08-13
Add support for logical assignment operators '&&=', '||=', and '??='.

mjambon:
I don't know if these features are used much... but they're easy to add so I thought it would be a good warm-up for me.

* ecmascript doc: https://tc39.es/ecma262/#sec-assignment-operators
* MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR_assignment


grammar.js
```diff
@@ -705,7 +705,8 @@ module.exports = grammar({
         $.identifier,
         $.parenthesized_expression,
       )),
-      choice('+=', '-=', '*=', '/=', '%=', '^=', '&=', '|=', '>>=', '>>>=', '<<=', '**='),
+      choice('+=', '-=', '*=', '/=', '%=', '^=', '&=', '|=', '>>=', '>>>=',
+             '<<=', '**=', '&&=', '||=', '??='),
```
mjambon, line 709: I like to cut lines at 80 characters. I could do differently if you want (or use an auto-format tool).

package.json
```diff
@@ -10,7 +10,8 @@
   "author": "Max Brunsfeld",
   "license": "MIT",
   "dependencies": {
-    "nan": "^2.12.1"
+    "nan": "^2.12.1",
+    "tree-sitter": "^0.16.1"
```
mjambon, line 14: I don't know why this was added or why it was not here before. I called `npm install tree-sitter` and `npm install tree-sitter-cli` beforehand.

Aerijo: IIRC `tree-sitter` is the NodeJS bindings. Shouldn't necessary for writing a grammar. It's there because part of `npm install <pack>` is to add `<pack>` to dependencies. Just run `npm install` to install all the existing package dependencies. I believe `tree-sitter-cli` is the package that provides the `tree-sitter` CLI program.

mjambon: I see Travis uses just `npm install` and `npm test` and this doesn't affect `package.json`.

maxbrunsfeld:
Thanks!

PR #139, opened 2020-10-03
Allow nested anonymous elements (React fragments)

mjambon:
Previously, `<>` was only allowed as the root element.

I have not tested this with typescript/tsx yet. Maybe it's fine to merge anyway. I'm working on typescript so I'll have to make sure very soon that it works there too.


maxbrunsfeld:
Thanks Martin!

PR #140, opened 2020-10-06
Tolerate unescaped newlines in string literals

mjambon:
This is useful for typescript and is considered not harmful in javascript given the scope of the tree-sitter project.

This PR is a follow-up to https://github.com/tree-sitter/tree-sitter-typescript/pull/91


mjambon:
This should be good to merge!

grammar.js
```diff
@@ -804,25 +805,34 @@ module.exports = grammar({
...
-          token.immediate(prec(PREC.STRING, /[^'\\\n]+|\\\r?\n/)),
+          token.immediate(prec(PREC.STRING, /[^'\\\n]+|\\?\r?\n/)),
           $.escape_sequence
         )),
         "'"
       )
     ),
 
+    string: $ => $._string,
+    jsx_string: $ => $._string,
```
maxbrunsfeld, line 834: I think there's no need to have two separate `string` and `jsx_string` rules.

Previously, I had been thinking of a different separation - in which, to be conservative, we would create a separate `jsx_string` rule that allowed unescaped newlines, and would only be used inside of JSX expressions. But even if we did that, I would have used `alias` to make sure that the `jsx_string` appeared as a normal `string` in the output. I don't think it's helpful to have a public-facing `jsx_string` rule, since it's just another node type that users need to remember to target for things like syntax highlighting.

But I think it's probably fine to just allow unescaped newlines in *all* strings, as you've done. I would just get rid of the `jsx_string` thing.

mjambon: Thanks. I will do this. (sorry about rushing this PR and then not reacting - it turns out this is a "hackathon week" at r2c and we're all doing different things this week)

mjambon: I just pushed the fix. There's now just one `string` rule that tolerates unescaped newlines.

mjambon: ping :)

mjambon:
Hi @maxbrunsfeld, we'd like to merge this. We've been using it in prod for a while and it works fine. I just updated the generated code for mergeability.

grammar.js
```diff
@@ -805,19 +805,25 @@ module.exports = grammar({
...
+    // This is legal in typescript as jsx/tsx attribute values (as of
+    // 2020), and perhaps will be valid in javascript as well in the
+    // future.
+    //
     string: $ => choice(
       seq(
         '"',
         repeat(choice(
-          token.immediate(prec(PREC.STRING, /[^"\\\n]+|\\\r?\n/)),
+          token.immediate(prec(PREC.STRING, /[^"\\\n]+|\\?\r?\n/)),
```
maxbrunsfeld, line 818: I think that if we're allowing newlines, you could simplify the regex so that newlines are not explicitly excluded on the left-hand-side of the OR operator. Then, the right-hand-side would only be for handling escaped newlines.

Something like `[^"\\]|\\\r?\n`

mjambon: Good point. Also, I'm now noticing that the regexp doesn't accept escaped quotes as in `"\""` or `'\''`. Looking into this.

mjambon: Nevermind, `$.escape_sequence` takes care of escaped quotes.

maxbrunsfeld:
Thanks!

grammar.js
mjambon: I simplified the regexp and added tests to cover my expectations. I also moved my original test for jsx attributes to `literals.txt` which seems more appropriate. `expressions.txt` still has some basic tests for strings and numbers which are probably fine to stay there.

mjambon:
Thank you Max!

PR #143, opened 2020-11-03
Designate "else"/"else if" branches inside of if statements

### LINKED ISSUE #142: Designate "else"/"else if" branches inside of if statements
Currently the `else` keyword is not captured inside of any node.

I want the whole expression, i.e. including the `else` keyword, to be scoped inside of a node, so that I can interact fully with it in my editor. Currently, for instance, it's not possible to cut, yank or delete a whole `else` branch due to this limitation.

resolritter:
closes  #142

code changes are included isolatedly in the first commit


resolritter:
Draft for now since I noticed one limitation: `else (block)` is included correctly, but `if else (block)` should also be nested inside of this new node. I'll figure out a fix for it.

grammar.js
```diff
@@ -239,14 +239,13 @@ module.exports = grammar({
       optional($._automatic_semicolon)
     )),
 
+    else_branch: $ => seq('else', $._statement),
```
maxbrunsfeld, line 242: Could you change this name to `else_clause`? I think that for these parts of statements, we usually use the term "clause" in this and other tree-sitter grammars. See for example `catch_clause`, `finally_clause` in this file.

maxbrunsfeld (commented):
👍  Your explanation in #142 makes sense, and I like this change.

I think the CI failure is because you generated the parser with `tree-sitter` 0.17.x, but the `package.json` file still has `tree-sitter-cli 0.16.x` in the `dependencies`. I think it would be ok to go ahead and upgrade `tree-sitter-cli` to the latest version as part of this PR. You just need to change the version number in `dependencies` to `0.17.3`.

grammar.js
resolritter: changed to `else_clause`

https://github.com/tree-sitter/tree-sitter-javascript/pull/143/commits/e03246ea76583c010b327a62e9f34168fccb9f73

resolritter:
> I think the CI failure is because you generated the parser with `tree-sitter` 0.17.x, but the `package.json` file still has `tree-sitter-cli 0.16.x` in the `dependencies`. I think it would be ok to go ahead and upgrade `tree-sitter-cli` to the latest version as part of this PR. You just need to change the version number in `dependencies` to `0.17.3`.

upgrade done in https://github.com/tree-sitter/tree-sitter-javascript/pull/143/commits/353c6feb6903d23e6e001554f4a181120ad99d38#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519, we'll see if CI passes...

resolritter:
ping @maxbrunsfeld I've made the requested changes. I can open another PR at tree-sitter-typescript for regenerating the parser after this lands.

maxbrunsfeld:
Thanks!

PR #146, opened 2020-11-25
Reuse LHS classes for "augmented_assignment_expression"

resolritter:
needed for https://github.com/tree-sitter/tree-sitter-typescript/pull/119


maxbrunsfeld:
Is a destructuring pattern allowed on the left-hand-side of an augmented assignment? It doesn't seem like it, based on my experiments, so it seems like this change would mildly increase the size of the parser without any benefit in terms of actual language coverage.

resolritter:
> Is a destructuring pattern allowed on the left-hand-side of an augmented assignment?

Doesn't seem to be. I've changed the approach, can you take a look again?

PR #149, opened 2021-02-06
Use empty_statement

icecream17:
I was using ```ctrl+f``` with ```";"```, when I saw that empty_statement and class_body had the same thing.

Very simple pr

### DIFF
--- a/src/grammar.json
+++ b/src/grammar.json
@@ -5521,8 +5521,8 @@
                     "type": "CHOICE",
                     "members": [
                       {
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
Hi, the `grammar.json` is automatically generated from `grammar.js`. Also, I don't think we want to parse that semicolon as an `empty_statement`. It's just part of a class body; we don't need to give it the name "empty statement" in the syntax tree. Thanks anyway!

PR #150, opened 2021-02-06
Add optional chaining to delimiters

icecream17:
Optional chaining and ...regular chaining should be in the same category


maxbrunsfeld:
Thanks!

PR #151, opened 2021-02-08
Fix "ambiguous argument 'head'" error in test script and make future errors fatal.

mjambon:
`git rev-parse head` was failing, causing a `git clone` every time. Taking it out of the `if` condition makes the shell script fail if this occurs again, instead of proceeding as previously.


maxbrunsfeld:
Thanks!

PR #152, opened 2021-02-10
Add more operators

icecream17:
Adds more operators

Justification for operators is that ```+= >> ||``` are already there.

Moves ```-- &&``` for consistency/visual purposes

(Ternary and keywords are not included as operators.)


.github/workflows/build.yml
```diff
@@ -1,6 +1,7 @@
 name: build
 
 on:
+  workflow_dispatch:
```
maxbrunsfeld, line 4: Can you remind me what this does?

maxbrunsfeld (commented):
Thank you!

.github/workflows/build.yml
icecream17: Oh, I was trying to test an addition to this pull request. It was on a pr of this... pull request. I added this while trying to do test the addition.
This allows for manual testing - but only if the default branch (```master```) has that too.

PR #155, opened 2021-02-25
New Patterns

BekaValentine:
This PR refactors the subgrammar for parameters and assignment LHSes to be a distinct set of patterns, instead of just a set of expressions. This makes it possible to refer to all variables that are used as binders, distinct from variables used as bound values, e.g.

`(pattern/identifier)` will match the parameter `x` in `(x) => x` but not the use of `x` in the body of the arrow function, while dually `(_expression/identifier)` will match the use but not the parameter.

This enables more precise syntax highlighting, but also is required for various precise codenav things to work.


grammar.js
```diff
@@ -719,8 +751,10 @@ module.exports = grammar({
     ),
 
     _destructuring_pattern: $ => choice(
-      alias($.object, $.object_pattern),
-      alias($.array, $.array_pattern)
+      // alias($.object, $.object_pattern),
+      $.object_pattern,
+      // alias($.array, $.array_pattern)
```
maxbrunsfeld, line 756: Let's 🔥  these comments.

maxbrunsfeld (commented):
This is great. I think that in some places in the Tree, `identifier` nodes that represent references will not be in `_expression` nodes. In particular, `member_expression` and `new_expression` seem to use `primary_expression` directly.

I made a branch (based off of this one) where I remove the `inline` for `primary_expression` (and drop the leading underscore in its name). You could check it out and merge it into your PR if you think it makes sense. There are no test changes, but `primary_expression` should now be usable as a supertype in queries (e.g. `(primary_expression/identifier)`.

BekaValentine:
ok, i've merged those branch changes into `new_patterns`

grammar.js
BekaValentine: done!

PR #160, opened 2021-03-10
Allow default export for any declaration

resolritter:
This change would cover the syntax described in https://github.com/tree-sitter/tree-sitter-typescript/pull/125#issue-540743941 without extra grammar additions on tree-sitter-typescript.


mjambon:
Hey @dcreager @maxbrunsfeld, could someone merge this please? We're eager to try this out and we can't do so easily because it makes us set up experimental forks for both tree-sitter-javascript and tree-sitter-typescript.

grammar.js
```diff
@@ -79,6 +80,9 @@ module.exports = grammar({
     [$.assignment_expression, $.object_assignment_pattern],
     [$.labeled_statement, $._property_name],
     [$.computed_property_name, $.array],
+    [$.export_clause, $.object, $.object_pattern],
+    [$._import_export_specifier, $.object, $.object_pattern],
+    [$.export_statement, $._property_name],
```
maxbrunsfeld, line 85: Could you report what error you get if you remove these conflicts? I'm assuming that, by allowing `export <expression>`, we're creating ambiguities between "export clauses (e.g. `export {foo as bar, baz}`) and exported object literals (e.g. `export {foo: bar}`).

I'm wondering if the latter case is really allowed, or if we should keep `default` as required when exporting an expression.

resolritter: - ["Shorthand pattern" is valid JavaScript](https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&spec=false&loose=false&code_lz=MYewdgzgLgBAhjAvDAzAKDQUwB4AcQBOsA3vDAL5A&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=false&targets=&version=7.13.11&externalPlugins=). I don't know which version added this syntax but it's also listed [in the MDN page](https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export).

- ["Shorthand pattern" is valid TypeScript](https://www.typescriptlang.org/play?#code/MYewdgzgLgBAhjAvDAjAWAFCYKYA8AOIATrAN7wwC+QA).

The way I comprehend it, this syntax is not for exporting *identifiers*, not object literals. It's sugar for `export const c = c` so you don't have to repeat the `c` twice. Fairly common to see in the wild.

From what I've tested in the links above, the "object version" of `export { foo: bar }` is not valid. 

---

The error I get by removing the conflicts is:

```

  'export'  '{'  ','  •  '}'  …

Possible interpretations:

  1:  'export'  '{'  (object_pattern_repeat1  ',')  •  '}'  …
  2:  'export'  '{'  (object_repeat1  ',')  •  '}'  …
  3:  'export'  (export_clause  '{'  ','  •  '}')
```

mjambon: I pushed the suggested fix, consisting in allowing expressions only after `export default`, not directly after `export`.

I made a new PR for this: https://github.com/tree-sitter/tree-sitter-javascript/pull/168

mjambon:
Closing this PR since it was extended by #168 which is now merged.

PR #166, opened 2021-04-07
fix: parse decorators on properties

alexlafroscia:
I work in a codebase that uses decorators quite a bit (Ember.js) and noticed that a decorator applied to a property didn't seem to work. The highlighting looks _okay_ if there's not initializer, but if the property is initialized, you get something like this:

<details>
  <summary>Screenshot of parsing error</summary>
  <p>

CleanShot 2021-04-07 at 16 25 47@2x

</p></details>

After checking out the grammar, it seems like decorators on properties weren't really supported at all!

Now it should parse (and highlight) them correctly.

<details>
  <summary>Screenshot of fixed parser</summary>
  <p>

CleanShot 2021-04-07 at 16 37 17@2x

</p></details>


alexlafroscia:
I also noticed that the `src/parser.c` file has diffs if you run `tree-sitter generate` on the `master` commit. Maybe someone slipped in a code change without re-generating it?

I have a GitHub Action set up to prevent this in a grammar that I maintain

https://github.com/alexlafroscia/tree-sitter-glimmer/blob/ca8d1533c91107039e7a3f84019378eefe357582/.github/workflows/verify.yml#L26-L51

If you're interested, I'd be happy to add something similar to the workflows for this repo so that that kind of problem is avoided here, too!

alexlafroscia:
I went and rebased this PR so that it separates out the parser diff from `master` and the actual changes I'm trying to make.

alexlafroscia:
It seems like the GitHub Actions configuration is maybe broken? CI didn't run at all here...

maxbrunsfeld:
Thanks!

PR #167, opened 2021-04-07
Get CI Running Again

alexlafroscia:
This takes a stab at getting CI running again, since it doesn't appear to be working right now


alexlafroscia:
Looks like that worked!

alexlafroscia:
I'm curious if it's even worth running the tests on different operating systems... is there an expectation that the behavior would differ between them? I would think that the only place something might break in an OS-specific way is with the `tree-sitter` CLI itself... 

It certainly doesn't hurt to keep it, but if it's not necessary, it's always nice to remove some overhead!

maxbrunsfeld:
Having at least a windows build and a linux build is good, because there is a little bit of hand-written C code, and it's good to compile it with both MSVC and either clang or gcc. I'm not sure we really need a macOS build, but I guess let's leave it for now. Thanks for the fix!

PR #168, opened 2021-04-16
Allow default export for any declaration (extension of #160)

mjambon:
Implements suggestion from pending PR https://github.com/tree-sitter/tree-sitter-javascript/pull/160#pullrequestreview-615790260


mjambon:
Hmm, CI is giving us an error that I don't get on my machine:
```js
export default function () { }
export default function name1() { }
```
is expected to be parsed into
```
  (export_statement
    (function (formal_parameters) (statement_block)))
  (export_statement
    (function_declaration (identifier) (formal_parameters) (statement_block)))
```
but CI finds a `function` instead of a `function_declaration`:
```
  (export_statement
    (function (formal_parameters) (statement_block)))
  (export_statement
    (function (identifier) (formal_parameters) (statement_block)))
```

1. I don't know what I did wrong or what CI is doing different than me. I'm using tree-sitter 0.19.4. This is strange.
2. Both interpretations seem valid since after `default` we allow a declaration (including `function_declaration`) or an expression (including `function`). I don't see a dynamic conflict being declared, and even if this was a case of dynamic conflict resolution, can it be non-deterministic?

mjambon:
I had forgotten the regenerate the parser after having undone/redone a bunch of times... This time it should work.

mjambon:
I just removed a conflict declaration that was left over by accident. The diffs are simple now.

mjambon:
ok, I'm merging this because there's nothing left that's controversial, and I'll close the original PR by @resolritter .

Issue #176, opened 2021-06-08
Support private class fields


zbuc:
The ECMAscript spec and most JavaScript engines [supports private class fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields) however tree-sitter can't parse them correctly.

PR #177, opened 2021-06-08
Support syntax for private properties and methods

zbuc:
Resolves https://github.com/tree-sitter/tree-sitter-javascript/issues/176

Checklist:
- [ ] All tests pass in CI.


zbuc:
I used the following test file to validate:

```javascript
class ClassWithPrivateField {
  #privateField
}

class ClassWithPrivateMethod {
  #privateMethod() {
    return 'hello world'
  }
}

class ClassWithPrivateStaticField {
  static #PRIVATE_STATIC_FIELD
}
```

aryx:
Could this be done by just adding an optional('#') in method and field definitions?
What does the official JS grammar say?

mjambon:
Hopefully you can avoid introducing the dynamic conflict by using `optional('#')` like @aryx suggests.

zbuc:
I updated the PR to define `PrivateIdentifier` and allow that for field and method members. I had to change the existing `.public_field_definition` to `.field_definition` because now it can be either public or private. That's what originally led me to define new `private_field_definition` and `private_method_definition` types

zbuc:
Unfortunately this is failing a test case where a private method is called:

```javascript
  class ClassWithPrivateMethod {
    publicMethod() {
      return this.#privateMethod()
    }
  
    #privateMethod() {
      return 'hello world'
    }
  }
```

zbuc:
OK, every test case from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields is passing now.

maxbrunsfeld (changes requested):
Your grammar changes look great AFAICT.

It looks like you didn't add any tests yet. Can you add tests in `test/corpus`?

grammar.js
```diff
@@ -723,7 +723,7 @@ module.exports = grammar({
     member_expression: $ => prec('member', seq(
       field('object', choice($.expression, $.primary_expression)),
       choice('.', '?.'),
-      field('property', alias($.identifier, $.property_identifier))
+      field('property', alias(choice($.identifier, $.private_identifier), $.property_identifier))
```
maxbrunsfeld, line 726: Maybe we should actually differentiate between private and public properties in the public API. What do you think?

If we do want that, I think we could:
* rename `private_identifier` to `private_property_identifier`
* in `_property_name`, move `private_property_identifier` *outside* the `alias()`
* similarly, change this line to:

```js
field('property', choice(
  $.private_property_identifier,
  alias($.identifier, $.property_identifier),
))
```

test/corpus/expressions.txt
```diff
@@ -343,6 +343,7 @@ class Foo {
   static one(a) { return a; };
   two(b) { return b; }
   three(c) { return c; }
+  #four(d) { return d; }
```
maxbrunsfeld, line 346: It doesn't look like there is a test for a private property in a *member expression* (which I see you added support for in the grammar). Is that indeed allowed?

If so, maybe you could update this line to say `{ return this.#d }`, or something like that.

maxbrunsfeld (commented):
I left a couple of questions.

grammar.js
zbuc: I think that makes a lot of sense. Updated.

maxbrunsfeld (approved):

PR #179, opened 2021-06-22
Add two issue templates: bug report and feature request

mjambon:
I'd like to add issue templates. This is intended to:
* minimize reports for the wrong project
* save time by having the reporter share what they already know

If we're happy with this, I'll make something similar for tree-sitter-typescript.

Checklist:
- [x] All tests pass in CI.


mjambon:
ping

aryx (approved):

Issue #181, opened 2021-06-30
Cannot assign to 'undefined'


mjambon:
The following piece of code is valid but it is parsed incorrectly:

```javascript
undefined = 42;
```

It's syntactically valid, although it doesn't make much sense since `undefined` is [read-only at least since ECMAScript 5](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined). Assigning to it has no effect. We have this in a test for semgrep, and it would be nice to not fail to parse this so we can pass the test.

The output of `tree-sitter parse` is the following:

```
$ tree-sitter parse <(echo 'undefined = 42;')
(program [0, 0] - [1, 0]
  (expression_statement [0, 0] - [0, 15]
    (undefined [0, 0] - [0, 9])
    (ERROR [0, 10] - [0, 14]
      (number [0, 12] - [0, 14]))))
/dev/fd/63	0 ms	(ERROR [0, 10] - [0, 14])
```
![image](https://user-images.githubusercontent.com/343265/123925557-e1d60a00-d93f-11eb-9406-4494973d2076.png)

In tree-sitter-typescript, we don't get an error but parsing is wrong (there shouldn't be a `non_null_expression` since indeed there's no `!`):

```
$ tree-sitter parse <(echo 'undefined = 42;')
(program [0, 0] - [1, 0]
  (expression_statement [0, 0] - [0, 15]
    (assignment_expression [0, 0] - [0, 14]
      left: (non_null_expression [0, 0] - [0, 9]
        (undefined [0, 0] - [0, 9]))
      right: (number [0, 12] - [0, 14]))))
/dev/fd/63	0 ms	(MISSING "!" [0, 9] - [0, 9])
```
![image](https://user-images.githubusercontent.com/343265/123925825-22ce1e80-d940-11eb-982c-e77f0b939b42.png)

In semgrep, we're switching to using the TSX parser (tree-sitter-typescript) for parsing javascript. This is how we discovered this issue. I'm hoping that fixing this in the javascript grammar will also fix it for typescript.

PR #183, opened 2021-06-30
Allow assignment to 'undefined'

mjambon:
This is an alternate solution to #181.

The previous solution was #182.

Checklist:
- [x] All tests pass in CI.


grammar.js
```diff
@@ -738,7 +738,7 @@ module.exports = grammar({
     _lhs_expression: $ => choice(
       $.member_expression,
       $.subscript_expression,
-      $.identifier,
+      $._identifier,
```
mjambon, line 741: This is what fixed the bug.

test/corpus/statements.txt
```diff
@@ -1,3 +1,37 @@
...
+var undefined = 42;
+const undefined = 42;
+let undefined = 42;
+
+---
+(program
+  (expression_statement (assignment_expression (undefined) (number)))
+  (variable_declaration (variable_declarator (identifier) (number)))
+  (lexical_declaration (variable_declarator (identifier) (number)))
+  (lexical_declaration (variable_declarator (identifier) (number))))
```
mjambon, line 33: Note that a simple assignment reports `undefined` as `undefined` but variable declarations (`var`, `const`, `let`) report it as an  `identifier`. This hasn't changed.

grammar.js
```diff
@@ -979,6 +979,13 @@ module.exports = grammar({
       ))
     },
 
+    // Matches the same set of tokens as $.identifier alone but reports
+    // 'undefined' differently.
```
aryx, line 983: maybe a comment to explain that even if undefined = xx does not make sense semantically, we want to accept it syntactically because ...

aryx (approved):

PR #196, opened 2021-08-21
Add some fields

FnControlOption:
Checklist:
- [x] All tests pass in CI.


FnControlOption:
More fields in progress

grammar.js
```diff
@@ -359,13 +359,13 @@ module.exports = grammar({
 
     return_statement: $ => seq(
       'return',
-      optional($._expressions),
+      optional(field('argument', $._expressions)),
```
pokey, line 362: I think I'd call this one `'value'`, as I believe it's more common to say "return value" instead of "return argument", and also seems consistent with other uses of "value" in this repo
```suggestion
      optional(field('value', $._expressions)),
```

PR #197, opened 2021-08-25
Fix rest pattern with assignment expression

FnControlOption:
Checklist:
- [x] All tests pass in CI.

Example:
```js
foo = [ ...[bar] = [baz] ]
```

which is equivalent to:
```js
foo = [ ...( [bar] = [baz] ) ]
```


maxbrunsfeld (approved):
This looks good. Thanks for adding good tests. I'll merge if you rebase it against master.

PR #198, opened 2021-08-25
Fix nested new expressions

FnControlOption:
Checklist:
- [x] All tests pass in CI.

Example:

```js
function Foo(x) {
  return function Bar(y) {
    this.baz = x + y
  }
}
let bar = new new Foo(1)(2)
console.log(bar.baz) // => 3
```


maxbrunsfeld:
Thanks!

PR #199, opened 2021-08-25
Fix for in/of statement with var initializer

FnControlOption:
Checklist:
- [x] All tests pass in CI.

Example:

```js
for (var foo = "bar" in []) {}
console.log(foo) // => bar
```

I added `$._initializer` to the `inline` array to resolve a conflict with `$.binary_expression`:

<details>

```

  'for'  '('  'var'  identifier  '='  expression  •  'in'  …

Possible interpretations:

  1:  'for'  '('  'var'  identifier  '='  (binary_expression  expression  •  'in'  expression)  (precedence: 'binary_relation', associativity: Left)
  2:  'for'  '('  'var'  identifier  (_initializer  '='  expression)  •  'in'  …

Possible resolutions:

  1:  Specify a higher precedence in `binary_expression` than in the other rules.
  2:  Specify a higher precedence in `_initializer` than in the other rules.
  3:  Specify a left or right associativity in `_initializer`
  4:  Add a conflict for these rules: `_initializer`, `binary_expression`
```

</details>


maxbrunsfeld:
I can see why you added that `inline`, and nice job finding a way to avoid that conflict. In this specific case though, I think that in the interest of reducing the binary size, I'd prefer to keep the conflict, and remove `_initializer` from the inlines.

Of course, the downside of the conflict is that there will be a very slight performance cost when parsing `for (var  x  =  y  in ...`, but the cost will be small, and this sequence doesn't occur very frequently. The advantage of removing the `inline` is that  the parser stays a bit smaller:

```diff
-#define STATE_COUNT 1628
-#define LARGE_STATE_COUNT 316
+#define STATE_COUNT 1562
+#define LARGE_STATE_COUNT 282
```

Are you ok with removing that `inline` and adding back the conflict?

maxbrunsfeld:
Thanks!

PR #201, opened 2021-09-08
fix: Issue on '?.' and ternary '?' + immediate dot started float number

ahlinc:
Checklist:
- [x] All tests pass in CI.

---
_Before:_
```
$ tree-sitter-0.20.0 parse <(echo "a?.1:.2")
(program [0, 0] - [1, 0]
  (ERROR [0, 0] - [0, 5]
    (identifier [0, 0] - [0, 1])
    (number [0, 3] - [0, 4]))
  (expression_statement [0, 5] - [0, 7]
    (number [0, 5] - [0, 7])))
0 ms	(ERROR [0, 0] - [0, 5])
```

_After:_
```
$ tree-sitter-0.20.0 parse <(echo "a?.1:.2")
(program [0, 0] - [1, 0]
  (expression_statement [0, 0] - [0, 7]
    (ternary_expression [0, 0] - [0, 7]
      condition: (identifier [0, 0] - [0, 1])
      consequence: (number [0, 2] - [0, 4])
      alternative: (number [0, 5] - [0, 7]))))
```

This PR fixes parsing of ternary expressions with an immediate dot float after question mark like: `a?.1:.2`.

Also the code in the `scanner.cc` was splited on separate functions.


test/corpus/expressions.txt
```diff
@@ -1021,120 +1447,184 @@ async = 0;
...
+================================================================================
 
 condition ? case1 : case2;
 
 x.y = some.condition ?
   some.case :
   some.other.case;
----
+
+a?.1:.2
```
ahlinc, line 1532: Here is an added test.

maxbrunsfeld (approved):

maxbrunsfeld:
Oh yeah, because of a conflict with the null coalescing member access operator.

Thanks for fixing this! And nice job organizing the external scanner code.

maxbrunsfeld:
It looks like there are some failures when running the parse example repos script. Could you investigate whether this change caused those?

ahlinc:
@maxbrunsfeld Thanks for the note, I've missed to support a case with leading spaces before the `?` in the external scanner. It's fixed now.

PR #202, opened 2021-09-17
0.20

Luni-4:
Checklist:
- [x] All tests pass in CI.
- [ ] The script/parse-examples passes without issues.
- [ ] Grammar rules have not been renamed unless absolutely necessary.
- [ ] The conflicts section hasn't grown too much.

This PR updates `tree-sitter-javascript` to 0.20

@maxbrunsfeld 

Can we release it on crates.io also? Thanks in advance! :)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "tree-sitter-javascript"
 description = "JavaScript grammar for the tree-sitter parsing library"
-version = "0.19.0"
+version = "0.20.0"
 authors = [
     "Max Brunsfeld <maxbrunsfeld@gmail.com>",
     "Douglas Creager <dcreager@dcreager.net>",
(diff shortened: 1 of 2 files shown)


aryx:
@mjambon which version of tree-sitter are we using currently in ocaml-tree-sitter-semgrep?

aryx (approved):

mjambon:
What's the motivation for this change, @Luni-4 ? My understanding is that the version constraint `^0.19.0` allows tree-sitter 0.20.0, but `^0.20.0` doesn't allow tree-sitter 0.19.0.

mjambon:
I wrote:
> My understanding is that the version constraint `^0.19.0` allows tree-sitter 0.20.0, but `^0.20.0` doesn't allow tree-sitter 0.19.0.

Sorry, I was wrong because `0.` is not a real major version. https://github.com/npm/node-semver#caret-ranges-123-025-004

PR #207, opened 2021-10-04
Loosen Rust crate's tree-sitter dependency

dcreager:
Nothing in the Rust binding depends on tree-sitter 0.20 in particular; it works just fine if you haven't updated from 0.19 yet.  (Other versions might work as well; these are the only two that I tested.)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ include = [
 path = "bindings/rust/lib.rs"
 
 [dependencies]
-tree-sitter = "0.20"
+tree-sitter = ">= 0.19, < 0.21"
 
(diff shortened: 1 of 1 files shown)


dcreager:
I've also backported this PR to a new `rust-0.19` branch that lets up publish a patch release of the 0.19 version of this crate.  (That doesn't include all of the recent grammar updates, which have been released as 0.20, and allows grammar consumers to update to tree-sitter 0.20 without having to update to the newest grammar as well.)

maxbrunsfeld (approved):

dcreager:
Published 0.19.1 and 0.20.0 with this update

PR #225, opened 2022-07-19
Add optional_chain named field

stackmystack:
Optional chaining property/array access and optional function
calls now produce an `optional_chain` named field.

When traversing a parsed program, once can retrieve the operator by
looking at the source code, indeed, but it seems _hacky_ (in a bad
sense).

Checklist:
- [ ] All tests pass in CI.


stackmystack:
Hello, is this the right place for these kinds of PRs?

mjambon:
Hi @stackmystack. Yes, this is the right place for such PRs. Sorry for the delays! For this particular issue, I'm not sure how `field()` is used in general as it's not a feature we use in semgrep. That's why I'd like to refer to @maxbrunsfeld.

aryx:
Also you get CI failures: "Error in query file "highlights.scm"

aryx (changes requested):
fix the CI failure (and do not click "CI test are passing" when they are actually not passing).

stackmystack:
Hello @aryx, I fixed the test-suite, it was working in the initial submission and got broken with a rebase which I didn't check.
All green now.

aryx (approved):
Why not.

PR #229, opened 2022-09-20
Add action to automatically publish crate

hendrikvanantwerpen:
Following other grammar repositories, add an action that automatically publishes the crate when a version tag is pushed.


hendrikvanantwerpen:
Already done in #240.

PR #237, opened 2023-01-24
Adds Blackbird queries

BekaValentine:
This PR adds queries to support GH code search.

Blackbird is the name of the GH code search tool that `tags.scm` was intended to support. For some reason the code search team was not using ts-internal tag queries and was instead using their own custom queries and we're now moving those queries upstream so that any service that wants to index name symbols in a way consistent with other services can rely on the tag queries.

Checklist:
- [ ] All tests pass in CI.
- [ ] The script/parse-examples passes without issues.
- [ ] Grammar rules have not been renamed unless absolutely necessary.
- [ ] The conflicts section hasn't grown too much.

### DIFF
--- a/queries/tags.scm
+++ b/queries/tags.scm
@@ -86,3 +86,14 @@
 
 (new_expression
   constructor: (_) @name) @reference.class
+
+(export_statement value: (assignment_expression left: (identifier) @name right: ([
+ (number)
+ (string)
+ (identifier)
+ (undefined)
+ (null)
+ (new_expression)
+ (binary_expression)
+ (call_expression)
+]))) @definition.constant


aryx:
Care to explain what blackbird queries are?

BekaValentine:
> Care to explain what blackbird queries are?

Sure! Blackbird is the name of the GH code search tool that the `tags.scm` was intended to support. For some reason the code search team was not using ts-internal tag queries and was instead using their own custom queries and we're now moving those queries upstream so that any service that wants to index name symbols in a way consistent with other services can rely on the tag queries.

aryx (approved):
Would be nice to add comment explaining the use of this export_statement
thing but whatever.

PR #238, opened 2023-01-26
Support import assertions

mskelton:
This adds support for [import assertions](https://github.com/tc39/proposal-import-assertions) which are at stage 3 and [experimental in Node](https://nodejs.org/api/esm.html#import-assertions).  Reading some other issues, I'm not sure this is desired at this stage, but I thought I would throw it out there to at least get thoughts on adding it.

Checklist:
- [ ] All tests pass in CI.


mjambon:
Hi. Sorry for the slow reaction time. I see that the [proposal](https://github.com/tc39/proposal-import-assertions) has been demoted from stage 3 to stage 2:

> :warning: This proposal has been demoted from Stage 3 to Stage 2 in January 2023 due to https://github.com/whatwg/html/issues/7233. TC39 is working on a solution, which will require relaxing the "assert only" semantics and potentially changing the syntax. Import assertions have already been shipped in some implementations, consider the current instability when using them.

I suggest we pause until we're more confident that the syntax will stick unless you can make a strong case for it.

mskelton:
@mjambon Yeah, I'm fine to close this with the change to stage 2.

bennypowers:
We're back to stage 3 with a new keyword and the feature is called "import attributes" now, instead of "import assertions"

https://github.com/tc39/proposal-import-attributes

Before:

```js
import manifest from './package.json' assert { type: 'json' };
```

After

```js
import manifest from './package.json' with { type: 'json' };
```

mskelton:
@bennypowers Yep, I opened #277 to support import attributes

PR #279, opened 2023-12-15
The semicolon should be interted before comment

cuyl:
- semicolon should be inserted before comment

Checklist:
- [x] All tests pass in CI.


amaanq:
I think I've managed this in my recent fixes, let me know if it's not okay or has any issues.

PR #284, opened 2024-01-04
Support HTML entities in JSX text/attributes

cpmsmith:
JSX text and attributes support HTML character references (a.k.a. entities), and don't support ECMAScript string escape sequences.

Although the [spec] calls it "historical" and threatens to change it, it _is_ in the spec, and the spec is pretty stable at this point.

In changing this, I landed back on an idea that @maxbrunsfeld suggested in a [PR review] some time ago: having separate `string` and `jsx_string` nodes, and aliasing `jsx_string` to `string` for consumers' convenience. At that time, having two different node types was deemed unnecessary, but this adds a second, more substantive difference between the two, so I've brought the idea back, and stopped allowing invalid newlines in JS string literals, which is invalid in both JS and TS.

[spec]: https://facebook.github.io/jsx/#sec-jsx-string-characters
[PR review]: https://github.com/tree-sitter/tree-sitter-javascript/pull/140#discussion_r503399344

## TL;DR
Here is some JSX highlighted in Neovim using tree-sitter:
image

And here it is in VSCode, not using tree-sitter:
image

VSCode, correctly, does not highlight the `\n` in the JSX attribute, and does highlight the two valid `&nbsp;`s, which tree-sitter-javascript doesn't currently parse. This PR fixes both things.

Checklist:
- [ ] All tests pass in CI. (awaiting approval)


amaanq:
I'm going to build on top of this with a fix for jsx vs js strings as well, thank you for the PR!

amaanq:
Hey again @cpmsmith, I cherry picked your changes onto #291, thank you for the PR!

PR #285, opened 2024-01-24
chore: split javascript and jsx parsers into separate dialects

### LINKED ISSUE #244: Split JSX into separate grammar?
As a user I want to be able to have separate settings for Javascript and JSX in my configuration. Currently support for JSX files (React.js extension for Javascript) is included in the Javascript grammar. However, Javascript and JSX are different languages, and I want to have separate per-filetype settings in my text editor.

I have recently added support for JSX in [nvim-ts-rainbow2](https://gitlab.com/HiPhish/nvim-ts-rainbow2) where I have two sets of queries for JSX files: one matches the angle brackets of tags, other other matches entire tags. This means a lot of copy-pasting between both queries.

If the two file types were separate I could make the JSX query inherit the Javascript query and add only the queries for the JSX extension. Example in TSX, note the first line which Neovim understands:

```; inherits: typescript

(jsx_element
  open_tag: (jsx_opening_element) @opening
  close_tag: (jsx_closing_element) @closing) @container
```

This is only possible because the https://github.com/tree-sitter/
[tree-sitter-typescript](https://github.com/tree-sitter/tree-sitter-typescript) repo splits the two languages this way as well. The result is a very terse query. If I want to change how I highlight tags I only need to write a query for that part and can piggy-back off the Typescript query for everything else.

calebdw:
Checklist:
- [x] All tests pass in CI.

Hello!

I just recently split the php grammar into separate dialects and decided to give this parser a whirl. This follows the same pattern as used by [tree-sitter-typescript](https://github.com/tree-sitter/tree-sitter-typescript) and [tree-sitter-php](https://github.com/tree-sitter/tree-sitter-php).

Note that the typescript grammar will need to be update after this is merged. I will try to create a sister branch on that repo in preparation for this.

Thanks!

Closes #244


calebdw:
Closing per https://github.com/tree-sitter/tree-sitter-javascript/issues/244#issuecomment-1908764548

PR #288, opened 2024-01-31
Misc bug fixes

### LINKED ISSUE #98: "Function calls" test case isn't valid JavaScript
In JavaScript, `function` at the start of a statement means the thing that comes after is parsed as a function statement, never an expression. So the following test case seems weird:

```
Function calls
============================================

x.someMethod(arg1, "arg2");
function(x, y) {

}(a, b);

---

(program
  (expression_statement (call_expression
    (member_expression (identifier) (property_identifier))
    (arguments (identifier) (string))))
  (expression_statement (call_expression
    (function
      (formal_parameters (identifier) (identifier))
      (statement_block))
    (arguments (identifier) (identifier)))))
```

I mean, the code doesn't parse in a real parser, so you could argue that anything that comes out of it is okay, but this is the only test that fails when I remove the (seemingly superfluous, because already part of `$._expression` via `$._constructable_expression`) `$.function` from the definition of `call_expression`, so I suspect it points at some confusion in the grammar definition.

```
    call_expression: $ => prec(PREC.CALL, seq(
      choice($._expression, $.super, $.function),
      choice($.arguments, $.template_string)
    )),
```

### LINKED ISSUE #188: Incorrect tree structure with coment starting in same line as closing bracket of a function

```javascript
class Docstring {
 
    /** DOCSTRING */
    foo1() {
      return;
     }         /*** WEIRD
     
     DOCSTRING
     ***/
    foo2() {
      return;
    }
}
```

Here's a link to the TypeScript Playground showing that the snippet above is valid JavaScript or TypeScript:

https://www.typescriptlang.org/play?filetype=js#code/MYGwhgzhAEAiD2wIBcBOBLAdgc2gbwChojpSB6AKgrgHkBhAZQBUAlASQDkBxaCsk0gDN48AIwAKAJT4BpaKgCmyAK6pMAblnQAvnL3kq1AOoBRNi1hatses3bcth-nuHwATFJn75S1RtnaBNpAA

The output of tree-sitter parsing is the following(using playground website: https://tree-sitter.github.io/tree-sitter/playground):
Screenshot from 2021-08-11 16-30-24

I've found an incorrect tree structure when there is a comment in the same line of a closing bracket from a function body.
As we can see, tree-sitter finds the "weird comment" inside the body of foo1 function even though this comment is written after the closing bracket of the body function. A semicolon or writting the comment below would fix the issue, but still this corner case could happen eventually since not everybody uses semicolon or maybe a misplaced line break accidentally falls in that place and breaks the tree structure.
Thanks!

### LINKED ISSUE #203: Incorrect parsing when keywords which are valid variable names are involved

```javascript
let + 5
```

[Here's a link to the TypeScript Playground showing that the snippet above is valid JavaScript or TypeScript](https://www.typescriptlang.org/play?alwaysStrict=false#code/G4QwTgBANgpgLhAvBArAbgFAdgg1KjIA)

The output of `tree-sitter` from the playground is the following:
```
program [0, 0] - [1, 0]
  ERROR [0, 0] - [0, 3]
  expression_statement [0, 4] - [0, 7]
    unary_expression [0, 4] - [0, 7]
      argument: number [0, 6] - [0, 7]
```

This should be an `expression_statement` of a `binary_expression` that adds the identifier `let` to the constant `5`.

---

```javascript
for (let in {});
```

[Here's a link to the TypeScript Playground showing that the snippet above is valid JavaScript or TypeScript](https://www.typescriptlang.org/play?alwaysStrict=false#code/GYewTgBAFANgpgFwgSwHYQN4F8CUBuAKCA)
Even TypeScript is wrong sometimes, you can run the code in the browser or Node.js to verify it's valid.

The output of `tree-sitter` from the playground is the following:
```
program [0, 0] - [1, 0]
  ERROR [0, 0] - [0, 16]
    lexical_declaration [0, 5] - [0, 16]
      ERROR [0, 9] - [0, 11]
        identifier [0, 9] - [0, 11]
      variable_declarator [0, 12] - [0, 14]
        name: object_pattern [0, 12] - [0, 14]
      ERROR [0, 14] - [0, 15]```
```

The expected result is a `for_in_statement`, like what we get when we change `let` to `letz` or something else.

---

Specifically `let` and `async` are valid variable names. `const`/`class` are not, which are handled correctly.

By the way [this test is from babel's test suite](https://github.com/babel/babel/blob/main/packages/babel-parser/test/fixtures/es2015/let/let-as-identifier-4/input.js), maybe you should run their tests as well.

### LINKED ISSUE #255: Comment not recognized in the middle of a ternary expression
This is a regression that was somehow introduced in e3a5149. This snippet parses correctly on its parent commit (2cc5803), but has been broken ever since.

The following piece of code is valid but it is parsed incorrectly, omitting the expected `comment` node:

```javascript
let foo = bar
  // this is a comment
  ? 'baz' : 'thud';
```

Here's a link to the TypeScript Playground showing that the snippet above is valid JavaScript or TypeScript:

https://www.typescriptlang.org/play?ssl=3&ssc=1&pln=3&pc=20#code/DYUwLgBAZg9jEF4JgE4FcQCgCQB6XyAFgJYDOEZEAhhAMYwC2DIAdmJhBAPwQDkARlQBevCAC4+YQmgAmvANxA

```
(program [0, 0] - [2, 19]
  (lexical_declaration [0, 0] - [2, 19]
    (variable_declarator [0, 4] - [2, 18]
      name: (identifier [0, 4] - [0, 7])
      value: (ternary_expression [0, 10] - [2, 18]
        condition: (identifier [0, 10] - [0, 13])
        consequence: (string [2, 4] - [2, 9]
          (string_fragment [2, 5] - [2, 8]))
        alternative: (string [2, 12] - [2, 18]
          (string_fragment [2, 13] - [2, 17]))))))
```

### LINKED ISSUE #268: Consider "un-aliasing" function named-node kind vs function token kind?
This isn't a bug I suppose but given that at least in the Rust `tree-sitter` bindings `Node::kind()` returns the "token value" for non-named nodes, there is a bit of a "namespace conflict" where actual `(function)` named nodes report their kind as `"function"` but then the non-named nodes for the `function` keyword (in function expressions/declarations) also reports its kind as `"function"`

So were one doing it from scratch, I'd probably suggest using eg `(function_expression)` instead of `(function)`

But I'm guessing "breaking" grammar changes should be minimized?

So I don't know if eg aliasing the `function` keyword token (`alias('function', 'function_keyword')`) would be less likely to break existing code?

The current situation should be work-around-able in most if not all situations, but it just "bit me" so figured I would raise the concern

Perhaps there is a deeper question there of whether named-node kinds should be sharing the same "namespace" as non-named-node kinds (at the Rust API level or wherever)?

### LINKED ISSUE #275: Expose template string chars as nodes?
It seems unexpected that if you traverse the "leaf" nodes of an AST with a template literal (that has some text content) there are no leaf nodes that contain that text content

For example for the source code `` `abc` `` the `template_string` node has child nodes for `` ` ``/`` ` `` but not `abc`

This appears to be asymmetrical wrt "regular" strings, which surface their contents as `string_fragment`'s

(and more generally I guess it violates some vague "invariant"/expectation I have that a node should either be a "leaf" node (ie have no child nodes) or else all of its non-whitespace characters should be encapsulated by some child node, no idea whether this is actually "a thing" or not)

My use case is that I basically want to compare whether two template strings are "identical" (as far as content/AST structure, so eg `` `abc${d}efg` `` and `` `abc${   d   }efg` `` should be considered "identical" even though they're not source-text identical) and this would presumably be way easier if the text content (here, `abc`/`efg`) were exposed as nodes (then I could just do leaf-node-by-leaf-node comparison)

I will open a PR that basically attempts to fix this by changing `_template_chars` -> `template_chars`, but I think I ran into something with highlighting/injection that makes it impossible to maintain existing behavior while adding this change

### LINKED ISSUE #283: Import statements with newlines in them not parsing correctly

```javascript
import { b } from 
'b'; /* comment */ import
 { a } from 'a';
```

Here's a link to the TypeScript Playground showing that the snippet above is valid JavaScript or TypeScript:

https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgIzgXzgMyhEcBQA5MoQNxwD0AVHAMa4gCmAdvFRXKJLPonAIbosOPIX5kgA

```
(program [0, 0] - [3, 0]
  (import_statement [0, 0] - [1, 4]
    (import_clause [0, 7] - [0, 12]
      (named_imports [0, 7] - [0, 12]
        (import_specifier [0, 9] - [0, 10]
          name: (identifier [0, 9] - [0, 10]))))
    source: (string [1, 0] - [1, 3]
      (string_fragment [1, 1] - [1, 2])))
  (comment [1, 5] - [1, 18])
  (expression_statement [1, 19] - [1, 25]
    (import [1, 19] - [1, 25]))
  (statement_block [2, 1] - [2, 6]
    (expression_statement [2, 3] - [2, 4]
      (identifier [2, 3] - [2, 4])))
  (expression_statement [2, 7] - [2, 11]
    (identifier [2, 7] - [2, 11]))
  (expression_statement [2, 12] - [2, 16]
    (string [2, 12] - [2, 15]
      (string_fragment [2, 13] - [2, 14]))))
/Users/jrosse/prj/tmp-js/tmp.js 0 ms    (MISSING ";" [2, 11] - [2, 11])
```

It looks like this should be parsing as two top-level `import_statement`'s but instead the second import statement is parsing as multiple top-level statements (I guess due to the newline in the middle of it)

### LINKED ISSUE #286: class A{;}

```javascript
class A{;}
```

Here's a link to the TypeScript Playground showing that the snippet above is valid JavaScript or TypeScript:

https://www.typescriptlang.org/play?#code/MYGwhgzhAECCDeBuAvkA

```
> tree-sitter-javascript@0.20.1 parse
> tree-sitter parse "1.js"

(program [0, 0] - [1, 0]
  (class_declaration [0, 0] - [0, 10]
    name: (identifier [0, 6] - [0, 7])
    body: (class_body [0, 7] - [0, 10]
      (ERROR [0, 8] - [0, 9]))))
1.js    0 ms    (ERROR [0, 8] - [0, 9])
```

### LINKED ISSUE #287: if (n-->0){} is parsed as JSX comment

```javascript
if (n-->0){}
```

Here's a link to the TypeScript Playground showing that the snippet above is valid JavaScript or TypeScript:

https://www.typescriptlang.org/play?#code/JYMwBAFAdgtDB8AGAlAbwL5A

```
> tree-sitter-javascript@0.20.1 parse
> tree-sitter parse 2.js

(program [0, 0] - [1, 0]
  (ERROR [0, 0] - [0, 5]
    (identifier [0, 4] - [0, 5]))
  (comment [0, 5] - [0, 12]))
2.js    0 ms    (ERROR [0, 0] - [0, 5])
```

amaanq:
Closes #98
Closes #188
Closes #203
Closes #255
Closes #268
Closes #275
Closes #283
Closes #286
Closes #287

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -500,7 +500,6 @@ module.exports = grammar({
       $.true,
       $.false,
       $.null,
-      $.import,
       $.object,
       $.array,
       $.function_expression,
@@ -767,7 +766,7 @@ module.exports = grammar({
 
     call_expression: $ => choice(
       prec('call', seq(
-        field('function', $.expression),
+        field('function', choice($.expression, $.import)),
         field('arguments', choice($.arguments, $.template_string)),
       )),
       prec('member', seq(
@@ -789,7 +788,7 @@ module.exports = grammar({
     )),
 
     member_expression: $ => prec('member', seq(
-      field('object', choice($.expression, $.primary_expression)),
+      field('object', choice($.expression, $.primary_expression, $.import)),
       choice('.', field('optional_chain', $.optional_chain)),
       field('property', choice(
         $.private_property_identifier,
--- a/grammar.js
+++ b/grammar.js
@@ -84,6 +84,7 @@ module.exports = grammar({
     [$.primary_expression, $.statement_block, 'object'],
     [$.import_statement, $.import],
     [$.export_statement, $.primary_expression],
+    [$.lexical_declaration, $.primary_expression],
   ],
 
   conflicts: $ => [
@@ -1204,6 +1205,7 @@ module.exports = grammar({
       'async',
       'static',
       'export',
+      'let',
     ),
 
     _semicolon: $ => choice($._automatic_semicolon, ';'),
--- a/grammar.js
+++ b/grammar.js
@@ -497,7 +497,7 @@ module.exports = grammar({
       $.import,
       $.object,
       $.array,
-      $.function,
+      $.function_expression,
       $.arrow_function,
       $.generator_function,
       $.class,
@@ -701,7 +701,7 @@ module.exports = grammar({
 
     class_heritage: $ => seq('extends', $.expression),
 
-    function: $ => prec('literal', seq(
+    function_expression: $ => prec('literal', seq(
       optional('async'),
       'function',
       field('name', optional($.identifier)),
--- a/grammar.js
+++ b/grammar.js
@@ -1120,6 +1120,7 @@ module.exports = grammar({
         seq(field('member', $.field_definition), $._semicolon),
         field('member', $.class_static_block),
         field('template', $.glimmer_template),
+        ';',
       )),
       '}',
     ),
--- a/grammar.js
+++ b/grammar.js
@@ -973,7 +973,7 @@ module.exports = grammar({
     template_string: $ => seq(
       '`',
       repeat(choice(
-        $._template_chars,
+        alias($._template_chars, $.string_fragment),
         $.escape_sequence,
         $.template_substitution,
       )),
--- a/grammar.js
+++ b/grammar.js
@@ -902,11 +902,7 @@ module.exports = grammar({
       ),
     )),
 
-    sequence_expression: $ => seq(
-      field('left', $.expression),
-      ',',
-      field('right', choice($.sequence_expression, $.expression)),
-    ),
+    sequence_expression: $ => prec.right(commaSep1($.expression)),
 
     //
     // Primitives
--- a/grammar.js
+++ b/grammar.js
@@ -17,10 +17,16 @@ module.exports = grammar({
     $._automatic_semicolon,
     $._template_chars,
     $._ternary_qmark,
+    $.html_comment,
+    '||',
+    // We use escape sequence to tell the scanner if we're currently inside a string or template string, in which case
+    // it should NOT parse html comments.
+    $.escape_sequence,
   ],
 
   extras: $ => [
     $.comment,
+    $.html_comment,
     /[\s\p{Zs}\uFEFF\u2028\u2029\u2060\u200B]/,
   ],
 
@@ -953,25 +959,16 @@ module.exports = grammar({
     )),
 
     // http://stackoverflow.com/questions/13014947/regex-to-match-a-c-style-multiline-comment/36328890#36328890
-    comment: _ => token(choice(
-      seq('//', /.*/),
-      seq(
-        '/*',
-        /[^*]*\*+([^/*][^*]*\*+)*/,
-        '/',
-      ),
-      // https://tc39.es/ecma262/#sec-html-like-comments
-      seq('<!--', /.*/),
-      // This allows code to exist before this token on the same line.
-      //
-      // Technically, --> is supposed to have nothing before it on the same line
-      // except for comments and whitespace, but that is difficult to express,
-      // and in general tree sitter grammars tend to prefer to be overly
-      // permissive anyway.
-      //
-      // This approach does not appear to cause problems in practice.
-      seq('-->', /.*/),
-    )),
+    comment: $ => choice(
+      token(choice(
+        seq('//', /.*/),
+        seq(
+          '/*',
+          /[^*]*\*+([^/*][^*]*\*+)*/,
+          '/',
+        ),
+      )),
+    ),
 
     template_string: $ => seq(
       '`',
--- a/queries/highlights.scm
+++ b/queries/highlights.scm
@@ -23,7 +23,7 @@
 ; Function and method definitions
 ;--------------------------------
 
-(function
+(function_expression
   name: (identifier) @function)
 (function_declaration
   name: (identifier) @function)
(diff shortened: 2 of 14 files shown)


vird:
Can I ask you release node-tree-sitter with up to date tree-sitter (at least 0.20.9)?
```
String.prototype.isUpperCase = function() {

```
(\n is important, and number of characters in property names are important)
works with tree-sitter-cli (ERROR is expected)
But with tree-sitter 0.20.6 is not working
```
#!/usr/bin/env node
Parser = require("tree-sitter")
ts_js  = require("tree-sitter-javascript")
parser = new Parser().setLanguage(ts_js)
code = "String.prototype.isUpperCase = function() {\n"
parser.parse(code)
```
Causes `Segmentation fault (core dumped)`

N.b. I tried to fix it with newer tree-sitter but no luck with install from github:tree-sitter/tree-sitter https://github.com/hu2prod/node-tree-sitter/blob/master/package.json#L31

amaanq:
I'll get to it when I can, catching up with a lot of tree-sitter stuff atm - so hopefully within a week or two

grammar.js
```diff
@@ -953,30 +955,21 @@ module.exports = grammar({
...
     // http://stackoverflow.com/questions/13014947/regex-to-match-a-c-style-multiline-comment/36328890#36328890
-    comment: _ => token(choice(
-      seq('//', /.*/),
-      seq(
-        '/*',
-        /[^*]*\*+([^/*][^*]*\*+)*/,
-        '/',
-      ),
-      // https://tc39.es/ecma262/#sec-html-like-comments
-      seq('<!--', /.*/),
```
aryx, line 964: what about those comments? We don't support them anymore? They were deprecated?

vird: I guess it was moved [here](https://github.com/tree-sitter/tree-sitter-javascript/pull/288/files#diff-bbcf6c785b007d9655b443abc204cc3cc32aa03cea190cb865fe44252dc6f1dfR223)

amaanq: Yeah it was moved to the external scanner as its own `html_comment` node because it was interfering with this code:
```
if(n-->0){}
```

PR #293, opened 2024-02-02
fix: tweak template literal injection

amaanq:
Fixes a core bug using this query for injection testing

### DIFF
--- a/queries/injections.scm
+++ b/queries/injections.scm
@@ -7,7 +7,10 @@
     (member_expression
       property: (property_identifier) @injection.language)
   ]
-  arguments: (template_string (string_fragment) @injection.content))
+  arguments: (template_string (string_fragment) @injection.content)
+  (#set! injection.combined)
+  (#set! injection.include-children))
+
 
 ; Parse regex syntax within regex literals
 


queries/injections.scm
```diff
@@ -7,7 +7,10 @@
     (member_expression
       property: (property_identifier) @injection.language)
   ]
-  arguments: (template_string (string_fragment) @injection.content))
+  arguments: (template_string (string_fragment) @injection.content)
+  (#set! injection.combined)
```
maxbrunsfeld, line 11: I think this use of `combined` will parse *all* template strings as one document, as opposed to a document per template string. Do we want that?

It used to work correctly before the template string fragment node got introduced. Now, I think we would need to use a repetition in the query to grab all of the string fragment nodes inside of a `template_string` in one match.

amaanq: I did try using a repetition at first in the query, but the test was not happy with it - it might have needed some tweak anyways, sorry about that. 

The test in question is here:
https://github.com/tree-sitter/tree-sitter/blob/ff89b539f276e2bd19248ffa462d8a426341993d/cli/src/tests/highlight_test.rs#L95

By just using the single string_fragment capture, this bit [here](https://github.com/tree-sitter/tree-sitter/blob/ff89b539f276e2bd19248ffa462d8a426341993d/cli/src/tests/highlight_test.rs#L121) changes:
from `("div", ["string", "tag"])` to  `("div", ["string"])`

Now if we use the one-or-more quantified capture `+`, we still don't get the correct result.

The split up html tag changes from this [here](https://github.com/tree-sitter/tree-sitter/blob/ff89b539f276e2bd19248ffa462d8a426341993d/cli/src/tests/highlight_test.rs#L120):
```
("</", ["string", "punctuation.bracket"]), 
("div", ["string", "tag"]), 
(">", ["string", "punctuation.bracket"])
```

to this:

`("</div>`", ["string"])`

I don't think that's what we want, but I also might have missed the obvious solution - sorry if I did

maxbrunsfeld: Ok, maybe `tree-sitter-highlight` doesn't deal properly with repeated captures in a match. I don't remember the state of that code, since it predated some improvements that have landed to the query engine. So this solution may be our best option for right now, even though ideally, I think we'd have separate injected parse trees per-template-string. Thanks for explaining. At some point, it would be good to understand why the repetition doesn't work, and fix anything as needed in `tree-sitter-highlight`.

amaanq: I agree, this solution didn't *feel* right to me - I can take a look at `tree-sitter-highlight` to see what might be causing the issue in the near future

PR #326, opened 2024-06-30
Fix: Comment optimization should not cause automatic-semi insert for single-line block

jackschu:
## Problem
Today the following piece of code is valid but it is parsed incorrectly:

```javascript
class C {
  name/*comment*/() {
  }
}
```

[Here's a link to the TypeScript Playground showing that the snippet above is valid JavaScript or TypeScript:
](https://www.typescriptlang.org/play/?#code/MYGwhgzhAEDC0G8BQ1oDswFsCmB6AVMAPaY5oAu+uAFAJSIrQC+STQA)

```
program [0, 0] - [4, 0]
  class_declaration [0, 0] - [3, 1]
    name: identifier [0, 6] - [0, 7]
    body: class_body [0, 8] - [3, 1]
      member: field_definition [1, 2] - [1, 6]
        property: property_identifier [1, 2] - [1, 6]
      comment [1, 6] - [1, 17]
      member: method_definition [1, 17] - [2, 3]
        name: MISSING property_identifier [1, 17] - [1, 17]
        parameters: formal_parameters [1, 17] - [1, 19]
        body: statement_block [1, 20] - [2, 3]
```

This is because `name/*comment*/() {` was being parsed as having an automatic semicolon between `name` and `()`, but this isn't right.

This was because there was an optimization that would cause automatic semicolons to be considered valid if a comment was started. This overlooked single line block comments. 

---
## Diff summary

Turns out the problem was really nuanced:

There are two callers of `scan_whitespace_and_comments`, 
1. an optimizing caller who wants to decide that having seen a comment is enough to consider auto-semi-insertion legal.
2. another who just wants to march the lexer forward maximally.

This PR basically gives the ability for the first of these to be invoked (ie because the automatic semicolon scanner saw a comment) and _not_ result in a decision on semicolon-legality in particularly the case where only a single-line-block-comment is seen.

It does this by 
1. changing `scan_whitespace_and_comments` to have an enum return value to indicate its three results (no-decision, reject, accept)
2. giving `scan_whitespace_and_comments` a consume:false mode where it stops as soon as it pops out of (a series of) block comments

Also added tests to the corpus

### DIFF
--- a/src/scanner.c
+++ b/src/scanner.c
@@ -48,9 +48,9 @@ static bool scan_template_chars(TSLexer *lexer) {
 }
 
 enum WhitespaceResult {
-    REJECT,
+    REJECT, // Semicolon is illegal, ie a syntax error occurred
     NO_NEWLINE, // Unclear if semicolon will be legal, continue
-    ACCEPT, // Semicolon is legal due to comment
+    ACCEPT, // Semicolon is legal, assuming a comment was encountered
 };
 
 /**
--- a/src/scanner.c
+++ b/src/scanner.c
@@ -47,16 +47,16 @@ static bool scan_template_chars(TSLexer *lexer) {
     }
 }
 
-enum WhitespaceResult {
+typedef enum {
     REJECT,     // Semicolon is illegal, ie a syntax error occurred
     NO_NEWLINE, // Unclear if semicolon will be legal, continue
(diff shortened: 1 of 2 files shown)


amaanq:
this is really good - thank you! just pushed a couple formatting changes and typedef'd the enum

PR #328, opened 2024-07-02
Fix: JSXText node trims whitespace too often

### LINKED ISSUE #325: JSXText node trims whitespace too often

```typescript
<div> <div/></div>
```

Here's a link to the TypeScript Playground showing that the snippet above is valid JavaScript or TypeScript:

https://www.typescriptlang.org/play/?#code/DwEwlgbgfABKkHorAeaQ

```
program [0, 0] - [1, 0]
  expression_statement [0, 0] - [0, 18]
    jsx_element [0, 0] - [0, 18]
      open_tag: jsx_opening_element [0, 0] - [0, 5]
        name: identifier [0, 1] - [0, 4]
      jsx_self_closing_element [0, 6] - [0, 12]
        name: identifier [0, 7] - [0, 10]
      close_tag: jsx_closing_element [0, 12] - [0, 18]
        name: identifier [0, 14] - [0, 17]
```

This is wrong because there is no jsx_text node
but as seen in the ts playground link `tsc` will output

```javascript
"use strict";
React.createElement("div", null,
    " ",
    React.createElement("div", null));
```

Note that " " will end up inside a dom node, so I think there should be a jsx_text node here.

This is not limited to an empty string ie
```typescript
<div> foo </div>
```
will result in a dom node with the text " foo " , so if I query for 'is there a jsx_text node with text === ' foo ' tree-sitter-javascript will tell me 'no' but i think the answer should be 'yes'

---
Advice / pointers
     
There is *a lot* of un-specified nuance here. Ie the whitespace rules are not part of the JSX spec (https://github.com/facebook/jsx/issues/143), but are implemented by React (https://github.com/facebook/jsx/issues/40)

Basically, i think the jsx_text node should grow to encapsulate the rules around whitespace rather than trimming always.

Relevant: #227

jackschu:
Closes #325 , see that issue for motivation details

I continued to uphold the request in #221 to not generate 'extra' 'jsx_text' nodes, this is _not_ what [babel](https://astexplorer.net/#/gist/e00813f5e66dd86747605a62dd106802/abd240ab5a0f5220ad543bad77d875f9bc2c7eba) or acorn do so I feel a bit iffy about this.

but completely-whitespace lines are nodes that are stripped by react's jsx_text emitter, so I think it could be OK for there to not be a syntax node here.
(See discussion https://github.com/facebook/react/pull/480#issuecomment-31296039)

---
I also updated some tests, ie 
`<Bar> <Foo /> </Bar>`

is meant to generate
```javascript
React.createElement(Bar, null,
    " ",
    React.createElement(Foo, null),
    " ");
```

So it makes sense that two jsx_text nodes are generated.

--- 
Side note: i got rid of the right branch of the `choice` that said `/\/\/[^\n]*/` which seemed to be trying to address would-be-comments. This is not effective ie 
```javascript
<div> // hi </div>
```
should treat ` // hi ` as a jsx_node, but it does not before or after PR (its an ERROR), this seems to be a separate issue
```javascript
React.createElement("div", null, " // hi ");
```

```javascript
<div> // hi  
</div>
```
Continues to work before and after this PR, but the inner jsx_text node spans 2 lines instead of one. I think this is preferred for tooling to consume because the emitted javascript includes trailing spaces after `hi` iff theres no newline before the `</div>`
(See discussion https://github.com/facebook/react/pull/480#issuecomment-31296039)


grammar.js
```diff
@@ -608,10 +608,11 @@ module.exports = grammar({
...
     ),
 
-    // Should not contain new lines and should not start or end with a space
     jsx_text: _ => choice(
-      /[^{}<>\n& ]([^{}<>\n&]*[^{}<>\n& ])?/,
-      /\/\/[^\n]*/,
+        // if there is a newline, only capture if there's non-whitespace-text
+        token.immediate(/[^{}<>&]*[^{}<>\n& ][^{}<>&]*/),
+        // whitespace between jsx_tags should be captured if there's no newline
+        token.immediate(/[^{}<>\n&]+/)
```
amaanq, line 615: spacing

```suggestion
      // if there is a newline, only capture if there's non-whitespace-text
      token.immediate(/[^{}<>&]*[^{}<>\n& ][^{}<>&]*/),
      // whitespace between jsx_tags should be captured if there's no newline
      token.immediate(/[^{}<>\n&]+/),
```

jackschu:
Ah thanks for the headsup and thanks for cleaning this up.

amaanq:
thank you very much for the PRs! sorry if I've been slow with reviewing, catching up on a lot :sweat_smile:

PR #331, opened 2024-07-07
Fix: expand meta_property to include import.meta

jackschu:
The estree spec includes `import.meta` as a new meta_property in es2020

https://github.com/estree/estree/blob/master/es2020.md#metaproperty

tree-sitter-javascript already considered `new.target` as a meta_property, which was inline inline with the spec at the time https://github.com/estree/estree/blob/master/es2015.md#metaproperty

So this diff brings us up-to-date and gives editors / tools the ability to treat these two meta properties similarly.

I also altered a test and re-generated with treesitter v0.22.6


amaanq:
ty!

PR #336, opened 2024-08-24
Fix: Operator precedence rules for tagged template functions

### LINKED ISSUE #334: apparent incorrect parse of `new` expression with a tagged template
It seems that the parse of a `new` expression combined with a tagged template is incorrect. Consider the snippet below.

```javascript
function f() {}
function i(template) { return f; }
new i`whatever`;
```

I expect this to be parsed as _new_expression_ where the constructor is a _member_expression_ consisting of _identifier_ `i` and _template_string_ `whatever`. However, this is being parsed as a _call_expression_ with a _new_expression_ and a _template_string_ as children:

```
     call_expression    "new i`whatever`"
       new_expression    "new i"
         new    "new"
         identifier    "i"
       template_string    "`whatever`"
         `    "`"
         string_fragment    "whatever"
         `    "`"
```

When the `new` expression contains arguments, the incorrectness becomes worse as two _call_expression_ are nested around the _new_expression_ and _template_string_, with arguments of the `new` expression being parsed as the arguments of the outer _call_expression_:

```
function g(v) {}
function j(template) { return g; }
new j`whatever`(1);
```

```
     call_expression    "new j`whatever`(1)"
       call_expression    "new j`whatever`"
         new_expression    "new j"
           new    "new"
           identifier    "j"
         template_string    "`whatever`"
           `    "`"
           string_fragment    "whatever"
           `    "`"
       arguments    "(1)"
         (    "("
         number    "1"
         )    ")"

```

As a reference to what I consider the correct parse for these 2 snippets, I'm attaching a screenshot from [AST explorer](https://astexplorer.net/)'s parse of it.
Screen Shot 2024-08-13 at 14 28 52

Perhaps this can be fixed by replacing `primary_expression` for `member_expression` in the [tree-sitter's grammar](https://github.com/tree-sitter/tree-sitter-javascript/blob/a92640f158c208ca3be28b560b838b2bd60e8eac/grammar.js#L821)? (Link to the corresponding [spec grammar](https://tc39.es/ecma262/#sec-left-hand-side-expressions).)

jackschu:
Closes #334 

## Background understanding
(FYI i'm using `'` as a backtick in in-line code blocks for convenience, whenever i use single-quote, i mean backtick)

Tagged templates have operator precedence ~16 according to MDN ([src](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#specifications:~:text=You%20can%20use%20any%20expression%20with%20precedence%20greater%20than%2016))

> The tag does not have to be a plain identifier. You can use any expression with [precedence](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_precedence#table) greater than 16, which includes [property access](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_accessors), function call, [new expression](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/new), or even another tagged template literal.

`new` without its `(..)` args is level 16

so `new foo'hi'` should result in `foo'hi'` being evaluated first 
then `new` is applied as if the result of `foo'hi'` is a classname

So the correct parse tree should be something like this
```
(new_expression [1, 0] - [1, 9]
      constructor: (call_expression [1, 4] - [1, 9]
        function: (identifier [1, 4] - [1, 5])
        arguments: (template_string [1, 5] - [1, 9]
          (string_fragment [1, 6] - [1, 8]))))
```

but `new()` with its args list is level 17
so `new foo()'hi'` should result in `new foo()` being evaluated first 
then `'hi'` being applied as it 'calls' the resulting class instance

So the correct parse tree should be something like this
```
expression_statement [0, 0] - [0, 12]
    (call_expression [0, 0] - [0, 11]
      function: (new_expression [0, 0] - [0, 7]
        constructor: (identifier [0, 4] - [0, 5])
        arguments: (arguments [0, 5] - [0, 7]))
      arguments: (template_string [0, 7] - [0, 11]
        (string_fragment [0, 8] - [0, 10])))
```

## Current Problem and fix
`new foo()'hi'`  is currently correct but
`new foo'hi'` results in

```
(expression_statement [0, 0] - [0, 11]
    (call_expression [0, 0] - [0, 11]
      function: (new_expression [0, 0] - [0, 7]
        constructor: (identifier [0, 4] - [0, 7])) // this is wrong!
      arguments: (template_string [0, 7] - [0, 11]
        (string_fragment [0, 8] - [0, 10])))
```

This occurs for two reasons, `call_expressions` require an `expression`, but `new_expressions` only allow `primary_expressions` so when the `identifier` `foo` is processed into a `primary_expression` it gets consumed by `new_expression` without conflict.

This diff changes `call_expressions` to be constructed with `choice(primary_expression, new_expression)`  in the case where a tagged-template is used. This is sane because of the precedence rules discussed earlier, ie `primary_expression` allows for `call_expression` `member_expression` (level 17), `parenthesized_expression` (level 18), and direct identifiers.

The only other allowed expression is `new_expression` (level 17)  so I added it to the choice block

With this change, we still don't get the desired result because 'new' primary_expression gets resolved as new_expression before we can make a call expression. So I add a new `template_call` precedence for this case.

Note that regular `call` precedences should still come after 'new' so that `new foo()` continues to parse correctly

cc @amaanq


amaanq:
awesome, thank you!

PR #341, opened 2024-10-26
feat!: remove glimmer template tags

amaanq:
There is now a glimmer parser that's been out for a while now, consumers interested in a glimmer parser should be using this. [link](https://github.com/NullVoxPopuli/tree-sitter-glimmer-javascript)


maxbrunsfeld:
Thanks for doing this

PR #354, opened 2025-05-27
Fix: correct ASI for arrow function followed by parenthesized expressions

jackschu:
### Observed
The following parses incorrectly
```js
()=>{}
(1)
```
Without this changeset, the grammar treats this as an arrow function that is being invoked.
```
(program [0, 0] - [1, 3]
  (expression_statement [0, 0] - [1, 3]
    (call_expression [0, 0] - [1, 3]
      function: (arrow_function [0, 0] - [0, 6]
        parameters: (formal_parameters [0, 0] - [0, 2])
        body: (statement_block [0, 4] - [0, 6]))
      arguments: (arguments [1, 0] - [1, 3]
        (number [1, 1] - [1, 2])))))
```
We can gut check that this is wrong by running the following in node
```js
() => {console.log('this never logs because the function is never invoked')}
(console.log('this logs'))
```
### Expected
Instead this should parse as an arrow function, an automatically injected semicolon, and a parenthesized number, eg
```
(program [0, 0] - [1, 3]
  (expression_statement [0, 0] - [0, 6]
    (arrow_function [0, 0] - [0, 6]
      parameters: (formal_parameters [0, 0] - [0, 2])
      body: (statement_block [0, 4] - [0, 6])))
  (expression_statement [1, 0] - [1, 3]
    (parenthesized_expression [1, 0] - [1, 3]
      (number [1, 1] - [1, 2]))))
```
### Background
Ill be using PascalCase to indicate spec nouns and snake_case to indicate tree-sitter-javascript nouns

This grammar considers two levels of expressions. `expression` and `primary_expression`. These both exist in the ecmascript spec, but are obviously not perfect representations of their highly-granular spec counterparts (this in intentional as the [tree-sitter docs](https://tree-sitter.github.io/tree-sitter/creating-parsers/3-writing-the-grammar.html#structuring-rules-well) point out). 

The spec considers a third kind of expression called [MemberExpression](https://tc39.es/ecma262/#prod-MemberExpression). This is just one level above PrimaryExpression in the spec and includes things that our `primary_expression` already includes (property access, subscript expression, `meta`, `super`, etc.).

### Changes
It turns out that there are _several_ issues in the way of handling this correctly, all of which I tackle in this PR. 

1. `call_expression`s currently accept `expression` as its body, [the spec](https://tc39.es/ecma262/#prod-CallExpression) has CallExpression accept basically exclusively MemberExpression as its body. This diff moves `call_expression` to accept exclusively `primary_expression`. On its own this doesn't clearly correct any issue directly, but it moves us towards a consistent view a top the spec and it'll be important for use in later steps.
2. Step 1. forces the issue that `new_expression` is considered to be considered a `primary_expression`, but the spec considers it to be a [MemberExpression](https://tc39.es/ecma262/#prod-MemberExpression). So productions like `new foo()()` won't parse but should. So I move `new_expression` to be considered a `primary_expression` (after all it seems like in spirit our `primary_expression` is a MemberExpressoin).
    - At this point, we can tackle the niche precedence rule where `new` with no args is lower precedence than call expressions but `new` with args is higher precedence ([src](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_precedence#:~:text=new%20with%20argument%20list,argument%20list%0Anew%20x)). 
    - Interestingly this properly places template call expressions properly at around precedence level 16 ([src](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#specifications:~:text=You%20can%20use%20any%20expression%20with%20precedence%20greater%20than%2016)) so we can drop `template_call` as a prec level (added in #336 )
1. arrow functions are currently considered a primary_expression, they are not a PrimaryExpression in the spec, and are not a MemberExpression either, so I move them to `expression`.


clason:
The test failures are due to the fact that you're generating the parser with ABI 15, which the bindings are not compatible with. Please use `tree-sitter g --abi=14` for this PR.

Also, please update the reference queries to the grammar change, if necessary.

clason:
Ugh, CI forces generation... 

Can you add `generate: false` after this line?
https://github.com/tree-sitter/tree-sitter-javascript/blob/6fbef40512dcd9f0a61ce03a4c9ae7597b36ab5c/.github/workflows/ci.yml#L43

(This should be a separate `ci(test): don't validate generation` commit _before_ your -- squashed! -- other commit.)

clason:
Yes, hence my asking you to disable the test (which you did, which is why CI passes now).

But please check that the queries are still compatbile!

clason:
As long as the old queries are still valid, that's ok. We've had PRs that change the grammar in a way that they break without thinking to adapt the reference queries here (because they use their own).

clason:
https://github.com/tree-sitter/tree-sitter-javascript/tree/master/queries

jackschu:
@amaanq could you take a look?

PR #378, opened 2026-03-07
Fix: allow reserved words as property identifiers after optional chaining

### LINKED ISSUE #377: Bug: Reserved words not recognized as property identifiers after optional chaining (`?.`)
The following pieces of code are valid but parsed incorrectly:

```javascript
a?.delete(b);
```

```javascript
myMap?.delete(key);
```

```javascript
a?.class;
a?.return;
```

**`a.delete(b);` parses correctly:**
```
(program [0, 0] - [1, 0]
  (expression_statement [0, 0] - [0, 12]
    (call_expression [0, 0] - [0, 11]
      function: (member_expression [0, 0] - [0, 8]
        object: (identifier [0, 0] - [0, 1])
        property: (property_identifier [0, 2] - [0, 8]))
      arguments: (arguments [0, 8] - [0, 11]
        (identifier [0, 9] - [0, 10])))))
```

**`a?.delete(b);` produces an ERROR node:**
```
(program [0, 0] - [1, 0]
  (expression_statement [0, 0] - [0, 13]
    (call_expression [0, 0] - [0, 12]
      (ERROR [0, 3] - [0, 9])
      arguments: (arguments [0, 9] - [0, 12]
        (identifier [0, 10] - [0, 11])))))
```

The `delete` keyword (and all other globally-reserved words) is correctly treated as a `property_identifier` after `.` but produces an `ERROR` node after `?.` (optional chaining).

All globally-reserved words are affected:
```javascript
a?.delete(b);  // ERROR
a?.class;      // ERROR
a?.return;     // ERROR
```

## Analysis

The `member_expression` rule in `grammar.js` (lines 884-891) uses `reserved('properties', ...)` to allow reserved words as property identifiers:

```javascript
member_expression: $ => prec('member', seq(
  field('object', choice($.expression, $.primary_expression, $.import)),
  choice('.', field('optional_chain', $.optional_chain)),
  field('property', choice(
    $.private_property_identifier,
    reserved('properties', alias($.identifier, $.property_identifier)),
  )),
)),
```

The `properties` reserved list is empty (line 72), meaning all globally-reserved words should be permitted in property position. This works correctly when the preceding token is `.`, but not when it is `?.` (`optional_chain`, which is an external token defined at line 855).

The `reserved` feature (introduced in tree-sitter v0.25) may not be applying its keyword-to-identifier demotion when the context involves external tokens like `optional_chain`. This could also be a tree-sitter core issue rather than a grammar issue.

## Real-World Impact

This affects real-world codebases significantly. `Map.prototype.delete()` and `Set.prototype.delete()` are commonly called via optional chaining (`map?.delete(key)`). Parsing the current [microsoft/vscode](https://github.com/microsoft/vscode) repository with tree-sitter-typescript (which inherits this grammar) produces 40+ parse errors, nearly all from `?.delete()` calls.

## Environment

- tree-sitter CLI: v0.26.6
- tree-sitter-javascript: v0.25.0
- Platform: macOS (darwin arm64), also reproduced on Linux and Windows via CI

rbonestell:
## Summary

Fixes #377.

Reserved words (`delete`, `class`, `return`, etc.) are correctly treated as `property_identifier` nodes after `.` but produce `ERROR` nodes after `?.` (optional chaining). This is because the `reserved('properties', ...)` context is not propagated through the `?.` token path — likely a tree-sitter core issue.

This PR works around it by explicitly listing all reserved words as valid alternatives in the `member_expression` property field via a new `_keyword_identifier` rule.

Additionally, this PR fixes JSX parsing of reserved/contextual keywords as tag names, attribute names, and member expression properties (a pre-existing issue). Credit to @SegaraRai for identifying and implementing the JSX fix ([commit](https://github.com/SegaraRai/tree-sitter-javascript/commit/e6d6e4287066543baf342d316b6bce8c9f1f2ae9)).

**Before:**
```
a?.delete(b);  →  ERROR
a?.class;      →  ERROR
a?.return;     →  ERROR
<class />      →  ERROR
<Foo.class />  →  ERROR
<a class="v"/> →  ERROR
```

**After:**
```
a?.delete(b);  →  (call_expression (member_expression … (optional_chain) (property_identifier)) (arguments …))
a?.class;      →  (member_expression … (optional_chain) (property_identifier))
a?.return;     →  (member_expression … (optional_chain) (property_identifier))
<class />      →  (jsx_self_closing_element (identifier))
<Foo.class />  →  (jsx_self_closing_element (member_expression (identifier) (property_identifier)))
<a class="v"/> →  (jsx_self_closing_element (identifier) (jsx_attribute (property_identifier) (string …)))
```

## Changes

- **grammar.js**: Extract `RESERVED_WORDS` and `CONTEXTUAL_KEYWORDS` as shared constants to eliminate duplication across `reserved.global`, `_reserved_identifier`, and the new `_keyword_identifier` rule. Add `_keyword_identifier` (inlined) to the `member_expression` property field alongside `$.identifier` inside the existing `reserved('properties', ...)` wrapper. Add `jsx_keyword_identifier` and `jsx_member_expression` rules to support reserved words in JSX contexts.
- **test/corpus/expressions.txt**: Add test case for optional chaining with reserved words.
- **test/corpus/jsx_identifiers.txt**: Add test cases for JSX reserved word identifiers (contributed by @SegaraRai).
- **package.json**: Add `tree-sitter` to `devDependencies` (required by `bindings/node/binding_test.js`).
- **.github/workflows/ci.yml**: Pin `node-version: 22` for Node binding tests (`tree-sitter@0.25.0` cannot compile with Node 24).
- **src/**: Regenerated via `tree-sitter generate`.

## Acknowledgments

Thanks to @SegaraRai for testing this branch and contributing the JSX reserved word fix ([#378 comment](https://github.com/tree-sitter/tree-sitter-javascript/pull/378#issuecomment-4149426829)).

## Notes

- All 118 corpus tests pass on all platforms.
- The `reserved('properties', ...)` wrapper is retained for the `.` path and forward compatibility — when tree-sitter core fixes the propagation issue, the explicit alternatives become redundant but harmless.
- The JSX fix introduces a dedicated `jsx_member_expression` rule (separate from `nested_identifier`) so that `_jsx_identifier` (which includes reserved words) is used for property resolution in JSX member expressions like `<Foo.class />`.


SegaraRai:
Thanks for putting this together. I tested this branch locally and noticed a follow-up issue on the JSX side:
reserved/contextual keywords no longer parse correctly as JSX tag and attribute names.

For example, cases like these no longer parse correctly in JSX:

- `<class />`
- `<svg:true />`
- `<Foo.class />`
- `<a class="v" for="x" const="y" await="z" />`

I put together a small follow-up fix on top of your branch here:
https://github.com/SegaraRai/tree-sitter-javascript/commit/e6d6e4287066543baf342d316b6bce8c9f1f2ae9

Please feel free to reuse or cherry-pick that commit if it is useful. I validated this locally against the affected JSX corpus cases, but I did not run the full CI matrix.

